PR #61, opened 2019-06-23
ref: remove jsonpath-rw which is slow

sbdchd:
jsonpath-rw seems to be the culprit clogging up the event loop (taking about 160ms).


kodiak/queries.py
```diff
@@ -512,29 +504,42 @@ def generate_jwt(self) -> str:
...
         pull_request["labels"] = labels
-        sha: typing.Optional[str] = get_value(
-            expr="repository.pullRequest.commits.nodes[0].commit.oid", data=data
-        )
+
+        try:
+            sha = data["repository"]["pullRequest"]["commits"]["nodes"][0]["commit"][
+                "oid"
+            ]
+        except (KeyError, TypeError):
```
chdsbd: should catch `IndexError` here

kodiak/queries.py
```diff
@@ -568,24 +576,33 @@ def find_branch_protection(
...
                 log.warning("Could not parse PRReview")
 
-        commit_status_dicts: typing.List[dict] = get_values(
-            expr="repository.pullRequest.commits.nodes[0].commit.status.contexts[*]",
-            data=data,
-        )
+        try:
+            commit_status_dicts: typing.List[dict] = data["repository"]["pullRequest"][
+                "commits"
+            ]["nodes"][0]["commit"]["status"]["contexts"]
```
chdsbd: ditto

kodiak/queries.py
```diff
@@ -594,36 +611,53 @@ def find_branch_protection(
...
-                data=data,
+        try:
+            valid_signature = bool(
+                data["repository"]["pullRequest"]["commits"]["nodes"][0]["commit"][
+                    "signature"
+                ]["isValid"]
             )
-            or False
-        )
+        except (KeyError, TypeError):
```
chdsbd: .

PR #71, opened 2019-07-16
fix(merge): ensure optimistic updates respect failing status checks

chdsbd:
If we have failing status checks we should raise NotQueueable instead of updating the PR


chdsbd:
My test here is incorrect and this actually behaves as expected.

PR #94, opened 2019-07-25
add: reporting of update branch failures

chdsbd:


kodiak/pull_request.py
```diff
@@ -254,15 +254,19 @@ def __repr__(self) -> str:
...
         event = await self.get_event()
         if event is None:
             self.log.warning("problem")
-            return
-        await self.client.merge_branch(
+            return None
+        res = await self.client.merge_branch(
             head=event.pull_request.baseRefName, base=event.pull_request.headRefName
         )
+        if res.status_code > 300:
```
sbdchd: feels like we should hook into the http client and have it automatically log non-ok errors

sbdchd: Different PR I suppose

PR #96, opened 2019-07-25
WIP: add: status message for CODEOWNERs

### LINKED ISSUE #87: Support CODEOWNERS for evaluating mergeability
We only display the default status message ("Merging blocked by branch protection settings") when a PR is blocked on a required code owner review. It would be nice to display a detailed message.

Related: #96

chdsbd:
We only display the default blocked status message when the PR requires a code owner to review. We should display a detailed status message about code owners.

TODO
- [x] fetch code owners file like we do kodiak.toml
- [x]  fetch files to check CODEOWNERS against
- [ ] find or build solution to parse code owners file
  + Chromium OWNERS: https://chromium.googlesource.com/chromium/tools/depot_tools/+/master/owners.py
  + Rust crates to write python bindings to: https://crates.io/search?q=Codeowners
  + JS solution: https://github.com/jamiebuilds/codeowners-utils
- [ ] evaluate code owners file against all files in a PR
- [ ] query if team or users have write permissions: https://github.community/t5/How-to-use-Git-and-GitHub/CODEOWNERS-works-with-users-but-not-teams/m-p/4991/highlight/true#M1612

NOTES
- I worry that some PRs could have huge amounts of files. So we may want to hard code some limit so we don't spend too much time looking at files. For intermediate amounts, we might be able to do something with the code owners file to return early, but I'm not sure how reliable that is to reduce the amount of work.
- use v3 api to fetch files as it has 300 per request vs 100 for v4: https://developer.github.com/v3/pulls/#list-pull-requests-files

Fixes #87


Timer:
What's the limitation here? You can access the file content and parse it.

chdsbd:
@Timer You're right. I think we can make code owners work, it's just going to require some effort to match the behavior of Github. I've updated the PR description.

chdsbd:
Closing this as described in #87

chdsbd:
I spoke too soon!

sanmai-NL:
@Timer: that's against the [information rights policy](https://kodiakhq.com/docs/permissions#what-data-is-read-by-kodiak), no? @chdsbd: I appreciate your efforts to keep efficient updates and owners working, but what's the cost/benefit ratio at this stage in Kodiak's evolution? Also seems like a feature sponsors might want to pay for, whereas basic correctness is a more pressing concern. Just my thoughts ...

chdsbd:
@sanmai-NL This PR predates that policy (and most everything else in this codebase 😄 ). This PR hasn't been touched since last August because it's low priority, but it is something I hope to get around to implementing.

chdsbd:
I looked into adding this feature again and I'm much more confident about it now. Fetching the first 100 files associated with a pull request does not appear to impact the API rate limit. I think 100 files is probably enough for most use cases.

However, it seems the main problem is team membership. I'm not sure of an efficient way to query team membership for code owners rules that require a team.

I think we'd need to listen for team membership webhook events and store our own copy of team membership information to query on. We'd probably want a cron job to periodically check team membership information to ensure we're okay when we miss events.

I think this PR will remain on ice for now. 😢

chdsbd:
#509 should resolve this issue. Follow along there if you're interested.

chdsbd:
Kodiak now supports CODEOWNERS with the release of #509 via https://github.com/chdsbd/kodiak/releases/tag/v0.30.0

PR #97, opened 2019-07-25
Fix typo in v1-opposite.toml

rgeraldporter:
Just a minor typo in a comment I saw. Thanks for this nifty app!


chdsbd:
Good catch. Thanks for the PR!

PR #110, opened 2019-07-25
fix: template of update branch failure

### LINKED ISSUE #101: Bad templating on failed to update message
See this checkrun for the bad template: https://github.com/zeit/next.js/pull/8020/checks?check_run_id=177364070

Need to make this an f-string: https://github.com/chdsbd/kodiak/blob/75843635ed3c21946103f388fd3545ae8acc4dfc/kodiak/queue.py

chdsbd:
Fixes #101

### DIFF
--- a/kodiak/queue.py
+++ b/kodiak/queue.py
@@ -194,7 +194,7 @@ async def repo_queue_consumer(
                         await asyncio.sleep(RETRY_RATE_SECONDS)
                     log.error("failed to update branch")
                     await pull_request.set_status(
-                        summary="🛑 could not update branch: {res}"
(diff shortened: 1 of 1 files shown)


kodiak/queue.py
```diff
@@ -194,7 +194,7 @@ def find_position(x: typing.Iterable[T], v: T) -> typing.Optional[int]:
                         await asyncio.sleep(RETRY_RATE_SECONDS)
                     log.error("failed to update branch")
                     await pull_request.set_status(
-                        summary="🛑 could not update branch: {res}"
+                        summary=f"🛑 could not update branch: {res}"
```
sbdchd: feels like there should be a lint for this

PR #134, opened 2019-07-27
fix: incorrect PR review validity check

### LINKED ISSUE #109: Fix evaluation of reviewer credibility
This check is inaccurate: https://github.com/chdsbd/kodiak/blob/5eceba2439209465ce5a27a4b7fefe24a7ba49c7/kodiak/evaluation.py#L177

Github docs: https://developer.github.com/v4/enum/commentauthorassociation/

NOTES
- It seems like `authorAssociation` is an insufficient check. We should instead be checking if the user has write access to the repository: Screen Shot 2019-07-26 at 11 32 19 PM
- I think we want to look at `assignableUsers`: https://developer.github.com/v4/object/repository/

Related: https://github.com/chdsbd/kodiak/issues/99

chdsbd:
Check that a user has write permission to make a meaningful review.

### NOTES
We need to find if a reviewer has write permission to a repo. This doesn't appear possible through the GraphQL api as the collaborators attribute only returns users and not their permissions. The v3 api may work.

The [collaborators list endpoint](https://developer.github.com/v3/repos/collaborators/#list-collaborators) appears bugged and doesn't show all users with write permissions to a repo. I've opened a ticket with Github, but, like many of my tickets, I haven't heard anything.
I've also posted on the community forum: https://github.community/t5/GitHub-API-Development-and/How-can-I-view-users-with-write-permissions-to-repo-through-the/td-p/28190

I think we need to do a query per reviewer on the [collaborators detail endpoint](https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level) to see if they have write permission to the repo. We could cache this if necessary.

Fixes #109


kodiak/queries.py
```diff
@@ -574,6 +588,47 @@ def __init__(self, *, owner: str, repo: str, installation_id: str):
...
+                headers=headers,
+            )
+        try:
+            res.raise_for_status()
+            return Permission(res.json()["permission"])
+        except (http.HTTPError, IndexError, TypeError, ValueError):
+            logger.exception("couldn't fetch permissions for username %r", username)
+            return Permission.NONE
+
+    async def get_reviewers_and_permissions(
```
sbdchd: test for this?

jukben:
❤️💪 Thanks fo this!

chdsbd:
@jukben You're welcome.

If you want to get notified when this is released to production you can subscribe to releases on the repo.
Screen Shot 2019-07-28 at 2 35 19 PM

PR #141, opened 2019-07-28
add: config to ignore select pending status checks

### LINKED ISSUE #118: Support configuring kodiak to not wait for certain checks
Sometimes a check will never succeed and will always stay in a pending state. The [wip app](https://github.com/apps/wip) is an example of this where while "WIP" is in the title of a PR, the WIP app will have a "pending" status check.

By default, kodiak would wait infinitely for this check to pass. The workaround we have now is the blacklist_title_regex, but we need something more robust to enable handling other checks.

We should support configuring checks that we don't want to wait to complete. Basically if these configured checks were in the "pending" state, we'd skip over merging the associated PR until the checks are in a success state.

Related #117

chdsbd:
Sometimes a check will never succeed and will always stay in a pending state. We should be able to configure these checks so that kodiak doesn't wait infinitely while these checks stay in the pending state.

Fixes #118

### DIFF
--- a/kodiak/evaluation.py
+++ b/kodiak/evaluation.py
@@ -202,10 +202,16 @@ def mergeable(
             passing_contexts: List[str] = []
             required = set(branch_protection.requiredStatusCheckContexts)
             for status_context in contexts:
+                # handle dont_wait_on_status_checks. We want to consider a
+                # status_check failed if it is incomplete and in the
+                # configuration.
                 if (
                     status_context.context in config.merge.dont_wait_on_status_checks
-                    or status_context.state in (StatusState.ERROR, StatusState.FAILURE)
+                    and status_context.state
+                    in (StatusState.EXPECTED, StatusState.PENDING)
                 ):
+                    failing_contexts.append(status_context.context)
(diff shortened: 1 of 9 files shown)


kodiak/evaluation.py
```diff
@@ -202,7 +202,10 @@ def mergeable(
             passing_contexts: List[str] = []
             required = set(branch_protection.requiredStatusCheckContexts)
             for status_context in contexts:
-                if status_context.state in (StatusState.ERROR, StatusState.FAILURE):
+                if (
+                    status_context.context in config.merge.dont_wait_on_status_checks
```
sbdchd: test for this?

kodiak/test_evaluation.py
```diff
@@ -481,6 +483,93 @@ def test_incomplete_checks(
     assert e.value.checks == {"wip-app"}
 
 
+def test_incomplete_checks_with_dont_wait_on_status_checks_check_run(
+    pull_request: PullRequest,
+    config: V1,
+    branch_protection: BranchProtectionRule,
+    review: PRReview,
+    check_run: CheckRun,
+) -> None:
```
sbdchd: doc strings?

PR #157, opened 2019-08-09
WIP: enable updating branches immediately

### LINKED ISSUE #120: Update branches immediately to match target
Hi 👋

Currently, PR branches are updated once all the other checks pass (approvals, label, ...).

I'd like to suggest updating the branch even when all those checks don't pass. The reason behind that is that we want to know that our branch is breaking as soon as possible. If we wait for all the other checks to pass, we might be losing time reviewing a branch that is broken once master in merged into it (and therefore will have to review it twice).

Do you think it is possible to implement this into Kodiak?

Thank you for your help 🙏

chdsbd:
Kodiak's been designed around updating and merging PRs efficiently. This change enables us to bypass that functionality and update PRs immediately when their target branches are updating.

TODO
- [ ] add configuration checks
- [ ] test feature

Fixes #120

### DIFF
--- a/kodiak/config.py
+++ b/kodiak/config.py
@@ -70,6 +70,7 @@ class Merge(BaseModel):
     # indefinite amount of time, like the wip-app checks or status checks
     # requiring manual approval.
     dont_wait_on_status_checks: List[str] = []
+    update_branch_immediately: bool = False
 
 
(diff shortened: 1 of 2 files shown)


chdsbd:
closing in favor of #159

PR #160, opened 2019-08-10
fix: docker caching

sbdchd:
Before we were installing deps after copying all of the code from the
repo into the container.

This breaks caching anytime the code changes.

Now the cache only breaks when either of the pyproject.toml and poetry.lock files change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,15 @@ WORKDIR /var/app
 
 COPY pyproject.toml poetry.lock /var/app/
 
+# install deps
 RUN poetry install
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,13 +1,17 @@
...
 
 RUN poetry config settings.virtualenvs.in-project true
+
+WORKDIR /var/app
+
+COPY pyproject.toml poetry.lock /var/app/
+
 RUN poetry install
 
+COPY . /var/app
```
chdsbd: you should install the CLI here

PR #162, opened 2019-08-11
WIP: fix: partial requiredStatusCheckContexts

chdsbd:
It appears that it's possible to configure requiredStatusCheckContexts with partial contexts. This change adds support for this behavior.

TODO
- replace set operations with substring matches. I think we want to check if the context starts with the required branch protection

### DIFF
--- a/kodiak/evaluation.py
+++ b/kodiak/evaluation.py
@@ -1,6 +1,6 @@
 import re
 from collections import defaultdict
-from typing import List, MutableMapping, Optional, Set
+from typing import Iterable, List, MutableMapping, Optional, Set
 
 import structlog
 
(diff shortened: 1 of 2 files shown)


chdsbd:
There seems to be something weird about this. I can only reproduce the behavior if the required status check is `continuous-integration/travis-ci` and the passing status check is like `continuous-integration/travis-ci/push`. If I try something like `continuous-integration/appveyor` with `continuous-integration/appveyor/push`, the behavior doesn't match.

I'm going to close this as I'm worried that if I try to match this behavior the implementation will be brittle. I haven't found documentation on this, so I'm not sure why this is happening.

PR #166, opened 2019-08-24
fix: travis-ci check compression

### LINKED ISSUE #163: PR doesnt get merged
sometimes a PR doesnt get merged by kodiak. in this cases kodiak states that it is waiting for 'travis ci' to get green.

but the ci checks already returned green.

this happened e.g. in the following PRs:
https://github.com/redaxo/redaxo/pull/2818
https://github.com/redaxo/redaxo/pull/2856

it works in 95% of our PRs as intended but sometimes doesnt... we werent able to identify why yet

see where all checks are already green but kodiak is still waiting for travis-ci:
grafik

we merged the PRs manually after waiting ~30min.

chdsbd:
This is blocked until I can find all the possible status checks that map to `continuous-integration/travis-ci`. I've opened an issue with GitHub, so hopefully they will have information.

fixes #163

### DIFF
--- a/kodiak/evaluation.py
+++ b/kodiak/evaluation.py
@@ -244,6 +244,20 @@ def mergeable(
             failing = set(failing_contexts)
             # we have failing statuses that are required
             failing_required_status_checks = failing & required
+            # GitHub has undocumented logic for travis-ci checks in GitHub
+            # branch protection rules. GitHub compresses
+            # "continuous-integration/travis-ci/{pr,pull}" to
+            # "continuous-integration/travis-ci". There is only special handling
+            # for these specific checks.
+            if "continuous-integration/travis-ci" in required:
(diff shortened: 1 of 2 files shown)


staabm:
As mentioned in the issue I try the more app 2 date travis integration.
In case this works it might be good enough to document this as a „known issue“ instead of complicating your code base ... its your call

chdsbd:
I talked with GitHub support a while back and they said there is handling just for the case of `continuous-integration/travis-ci/{pr,push} => continuous-integration/travis-ci`. That said I don't think it's worth adding this complication when the check is deprecated by travis-ci.

chdsbd:
Looking at the logs on Kodiak this does appear to affect people regularly, so I'm going to fix it.

Issue #174, opened 2019-09-10
use update_branch_immediately=true with require_automerge_label=true


clxmstaab:
we use the following kodiak config

```
# .kodiak.toml
version = 1

[merge]
method = "squash"
# once a PR is merged into master, delete the branch
delete_branch_on_merge = true
# label to use to enable Kodiak to merge a PR
automerge_label = "automerge" # default: "automerge"
# require that the automerge label be set for Kodiak to merge a PR. if you
# disable this Kodiak will immediately attempt to merge every PR you create
require_automerge_label = true
# if a PR is passing all checks and is able to be merged, merge it without
# placing it in the queue. This will introduce some unfairness where those
# waiting in the queue the longest will not be served first.
prioritize_ready_to_merge = true
# immediately update a PR whenever the target updates. If enabled, Kodiak will
# not be able to efficiently update PRs. Any time the target of a PR updates,
# the PR will update.
update_branch_immediately = true

[merge.message]
title = "pull_request_title"
body = "github_default"
include_pr_number = true
strip_html_comments = true # remove html comments to auto remove PR templates
```

and we have the following use-case:
we would like kodiak to automatically merge changes from the target branch into the PR, but do not yet merge them. therefore we use a `WIP:` title-prefix and mark PRs as soon as we want kodiak to update them with the merge-label.

it seems kodiak does not start auto-merging in changes from the target branch as long the title uses `WIP:` prefix... any chance we can get the desired behaviour?

sbdchd:
Hmm, ‘update_branch_immediately’ should update the PR when the target branch changes.

My first thought is the title blacklist regex is taking precedence over the ‘update_branch_immediately’ feature.

Sounds like a bug in Kodiak

chdsbd:
Okay, I've looked into #175 and I think we should actually add another configuration option for this behavior. Right now `update_branch_immediately` affects PRs that are mergeable and need an update. I think the behavior you want is to update a PR even if it is marked as "not mergeable" by having WIP in the title.

Would you want kodiak to update a branch even if it has failing status checks, or should this configuration option be limited to WIP regex?

I was thinking we could add a single flag to cover the case of the WIP regex, but we may want to make it a list of options too.

```toml
update_branch_immediately_ignore_wip_regex = true
# or
update_branch_immediately_ignore_blocker = ["wip_regex", "status_checks"]
```

I'm curious to hear your thoughts.

clxmstaab:
[...]

PR #187, opened 2019-11-20
Add ability to do smart updating without merging 

alex-sf:
## Why

In our workflow we need the ability to update a PR without merging it. This is because Kodiak does not support signed commits and as such we would like to keep our branches up to date once they are mergeable but wait to be manual merged. 

## What 

I think you were really close to this feature already. 

The first thing is your `update_pr_immediately_if_configured` already nicely waits for the PR to mergeable before it starts updating:

`update_pr_immediately_if_configured` only triggers if two things are true:
	1) `update_branch_immediately` flag is `true` 
	2) `m_res == MergeabilityResponse.NEEDS_UPDATE`
`1)` will protect this function from running for individuals who do not want this option. And `2)` actually already waits for the "smart updating" cycle because `MergeabilityResponse` prioritizes branch protection checks over `NEEDS_UPDATE` which means the branch will only be flagged as needing updating once it is ready to merge.

So the only thing I added was a new variable `do_not_merge` which will ensure that we always short circuit this function and do not add it to the merging queue. 

This will allow you to combine the flags `do_not_merge` and `update_pr_immediately_if_configured` to use kodiak to do smart auto updating in your PRs without merging. 

## Notes

If this idea looks like it's along the right path I can add more documentation/information to the README.

It looks like there are some other places Im missing the config variable. I can add those in as well shortly.

### DIFF
--- a/kodiak/queue.py
+++ b/kodiak/queue.py
@@ -125,8 +125,14 @@ async def process_webhook_event(
             isinstance(event.config, V1)
             and event.config.merge.do_not_merge
         ):
+            # we duplicate the status messages found in the mergeability
+            # function here because status messages for WAIT and NEEDS_UPDATE
+            # are only set when Kodiak hits the merging logic.
+            if m_res == MergeabilityResponse.WAIT:
+                await pull_request.set_status(summary="⌛️ waiting for checks")
+            if m_res in (MergeabilityResponse.OK, MergeabilityResponse.SKIPPABLE_CHECKS):
+                await pull_request.set_status(summary="✅ okay to merge")
             log.debug("skipping merging for PR because `merge.do_not_merge` is configured.")
-            await pull_request.set_status(summary="✅ okay to merge")
             return
 
         if (
--- a/kodiak/queue.py
+++ b/kodiak/queue.py
@@ -65,9 +65,13 @@ async def update_pr_immediately_if_configured(
         and event.config.merge.update_branch_immediately
     ):
         log.info("updating pull request")
+        await pull_request.set_status(summary="🔄 attempting to update pull request")
         if not await update_pr_with_retry(pull_request):
             log.error("failed to update branch")
             await pull_request.set_status(summary="🛑 could not update branch")
+        # we don't need to overwrite the "attempting to update..." message here
+        # because it will be updated by either the merge queue logic or the
+        # `merge.do_not_merge` logic.
 
 
 async def process_webhook_event(
@@ -121,6 +125,8 @@ async def process_webhook_event(
             isinstance(event.config, V1)
             and event.config.merge.do_not_merge
         ):
+            log.debug("skipping merging for PR because `merge.do_not_merge` is configured.")
+            await pull_request.set_status(summary="✅ okay to merge")
             return
 
         if (
--- a/kodiak/queue.py
+++ b/kodiak/queue.py
@@ -117,7 +117,10 @@ async def process_webhook_event(
         ):
             raise Exception("Unknown MergeabilityResponse")
 
-        if event.config.merge.do_not_merge:
+        if (
+            isinstance(event.config, V1)
+            and event.config.merge.do_not_merge
+        ):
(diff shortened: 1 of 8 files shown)


chdsbd:
@alex-sf Thanks for the PR, I'll have time to review it tonight.

alex-sf:
We can also rename the `do_not_merge` to something more explicit like `only_update_and_do_not_merge` and then have the `update_pr_immediately_if_configured` look for that flag as well 
```
async def update_pr_immediately_if_configured(
    m_res: MergeabilityResponse,
    event: EventInfoResponse,
    pull_request: PR,
    log: structlog.BoundLogger,
) -> None:
    if (
        m_res == MergeabilityResponse.NEEDS_UPDATE
        and isinstance(event.config, V1)
        and (event.config.merge.update_branch_immediately or event.config.merge.only_update_and_do_not_merge)
    ):
        log.info("updating pull request")
        if not await update_pr_with_retry(pull_request):
            log.error("failed to update branch")
            await pull_request.set_status(summary="🛑 could not update branch")
```
so we don't have to combine two flags together.

alex-sf:
Also the linter is outputting:
```
+ [[ -n true ]]
+ poetry run black --check .
would reformat /root/project/kodiak/config.py
All done! 💥 💔 💥
1 file would be reformatted, 28 files would be left unchanged.
Exited with code 1
CircleCI received exit code 1
```
which is not super helpful as to what needs to be reformatted in the `config` file. So Im not sure what to change in there.

chdsbd:
@alex-sf If you run the s/lint command locally it will format the code for you automatically.

staabm:
Not sure I get what this PR is about.

 Could you describe what the difference is to the WIP label and the issue reported in https://github.com/chdsbd/kodiak/issues/174

alex-sf:
@chdsbd I've never actually done and python so dont have it setup on my system to run the linter locally. If you can pull this change down, run the linter and push up the change when you review this pr that would be amazing! 

______

@staabm 
> Could you describe what the difference is to the WIP label and the issue reported in #174

That issue sounds like it's trying to allow updating while the PR is in WIP state. But ultimately would still auto merge. That is a completely different flow. This is blocking all merging and only auto update when the PR is fully mergeable. So it would maintain the same updating flow as the `update_branch_immediately` flag but would add a hard stop to merging so you can ensure that it will never try and auto merge.

chdsbd:
I tested locally and the change works. I've added some status messages so there's an indication that Kodiak is running.

The status messages are a bit hacky because some of the existing status messages are only triggered in the merging logic, but it works.

Here are some screenshots.

waiting for status check
image

failed status check
image

success
image

Feel free to change the status messages.

kodiak/queue.py
```diff
@@ -117,6 +121,22 @@ def find_position(x: typing.Iterable[T], v: T) -> typing.Optional[int]:
         ):
             raise Exception("Unknown MergeabilityResponse")
 
+        if isinstance(event.config, V1) and event.config.merge.do_not_merge:
+            # we duplicate the status messages found in the mergeability
+            # function here because status messages for WAIT and NEEDS_UPDATE
+            # are only set when Kodiak hits the merging logic.
+            if m_res == MergeabilityResponse.WAIT:
+                await pull_request.set_status(summary="⌛️ waiting for checks")
+            if m_res in (
```
sbdchd: just a style thing but I prefer this to be a set rather than a tuple

alex-sf:
@chdsbd @sbdchd thanks guys for all the work getting this over the line! When can we expect this to be released? We would like to integrate this into our companies work flow asap since waiting on PRs has been a huge issue for awhile now.

chdsbd:
@alex-sf I'm going to try deploying this tonight after work.

chdsbd:
Thanks for the contribution! :D

chdsbd:
I've deployed this change and it appears to work in my testing. Let me know if you run into any issues.

PR #202, opened 2019-12-20
enable updating forks

### LINKED ISSUE #104: support updating pull requests made from forks
Github's API doesn't support updating forks: https://developer.github.com/v3/repos/merging/#perform-a-merge

For reference, if it did, we would need to update this code to use the head/repo information of the fork:
https://github.com/chdsbd/kodiak/blob/7375b45e9d095d7532c9401792a15427ab428993/kodiak/pull_request.py#L296-L299

To work around this API limitation I thought we might be able to [clone the fork](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#http-based-git-access-by-an-installation) and do the update ourselves, but github app permissions are such that even if a PR is marked "Allow edits from maintainers.", our github app installation doesn't have permission to download that fork.

I believe we cannot do anything to implement this feature until github changes their API.

Related #100

chdsbd:
Use pull request update-branch api instead of merge api to update branches. This new api allows us to update branches of forks, which wasn't possible with the merges api because of GitHub permissions.

fixes #104

### DIFF
--- a/kodiak/messages.py
+++ b/kodiak/messages.py
@@ -1,6 +1,0 @@
-FORKS_CANNOT_BE_UPDATED = """\
-The [Github merging API](https://developer.github.com/v3/repos/merging/) only supports updating branches within a repository, not across forks. While Kodiak can still merge pull requests created from forked repositories, it cannot automatically update them.
-
-Please see [kodiak#104](https://github.com/chdsbd/kodiak/issues/104) for the most recent information on this limitation.
-
-"""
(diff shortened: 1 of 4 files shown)


KaiSzuttor:
I just tested it and it seems to work for forks. see https://github.com/KaiSzuttor/espresso/pull/40

chdsbd:
Thanks for testing. I’m going to try to merge this and release it tonight.

chdsbd:
I've tested this and I'm comfortable releasing this change.

PR #210, opened 2019-12-30
ref: GitHub StatusEvent parsing to be more lenient around URLs

sbdchd:
Previously we were using `pydantic`'s `UrlStr` which includes a regex
check that was failing in some cases.

We don't store the events so I'm not entirely sure the value returned
by GitHub so I copied and updated a fixture to produce a similar error
to the one in Sentry.

SENTRY: KODIAK-52


kodiak/github/fixtures/status_event/missing_target_url.json
```diff
@@ -0,0 +1,196 @@
+{
+  "id": 6671486559,
+  "sha": "736e0e8d7cfd6cc3ac43357f61ea0419debd57fa",
+  "name": "recipeyak/recipeyak",
+  "target_url": "example.com",
```
sbdchd: the tricky url :D

PR #219, opened 2020-01-06
fix: pydantic warnings about email-validator

chdsbd:
Pydantic complains that email-validator is not installed by printing to console, even though we don't use it. Anyway, adding the package will remove this annoyance.

```
email-validator not installed, email fields will be treated as str.
To install, run: pip install email-validator
```


chdsbd:
closing in favor of #233

PR #233, opened 2020-01-12
dep: upgrade fastapi from 0.33.0 to 0.46.0

chdsbd:
This change gives us pydantic 1.3, which fixes the email-validator warning. This change also updates starlette to 0.12.9.

There have been breaking changes in pydantic, specifically booleans are more strict and do not parse None as false. This hit one of our tests, but I'm worried that there may be other events that are incorrect.

https://github.com/samuelcolvin/pydantic/releases

This PR needs rigorous testing to ensure event parsing doesn't change.

closes #219


chdsbd:
I'll open another PR with this change.

PR #243, opened 2020-01-14
wip: ref: simplify event parsing logic

chdsbd:
This github event logic was originally intended to be a more general library, but now it's more of a liability because the schemas don't match precisely.

This change simplifies the schemas to their minimum to reduce transient parsing errors. This change also removes the type hint introspection for defining event handlers.


chdsbd:
I might do this later but I'm going to hold off for now

PR #247, opened 2020-01-17
fix: PRs not reevaluating on base branch update

### LINKED ISSUE #244: updates to master don't trigger updates on dependent PRs
When new commits are added to the target branch, the open PRs against that branch are not updated, even though `update.always` is configured.
report: https://github.com/chdsbd/kodiak/issues/174#issuecomment-574077658

I tried to reproduce this and I found that when CircleCI is configured, new commits on the target branch trigger a `status` webhook event for that branch, which triggers Kodiak to enqueue events for all the related PRs. I think if CircleCI used the newer GitHub Checks API, this issue would appear.

https://github.com/chdsbd/kodiak/blob/e1f9680f3cd136bd67ffb1a0648c2b756cd98229/kodiak/main.py#L96-L109

I think the fix is to accept a `push` webhook event and use that trigger PR updates.

chdsbd:
We now reevaluate all dependent PRs when their base branch updates (sends push webhook event).

Also remove unnecessary `res_json=res.json()` tags in logging because we cover those with `add_request_info_processor`.

fixes #244

### DIFF
--- a/kodiak/main.py
+++ b/kodiak/main.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import sys
+from typing import Optional
 
 import sentry_sdk
 import structlog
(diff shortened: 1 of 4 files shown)


kodiak/main.py
```diff
@@ -131,6 +132,43 @@
...
+    """
+    Extract the branch name from the ref
+    """
+    if raw_ref.startswith("refs/heads/"):
+        return raw_ref.split("refs/heads/", 1)[1]
+    return None
+
+
+@webhook()
+async def push(push_event: events.PushEvent) -> None:
```
sbdchd: I think it might be good to document either here with a doc string or somewhere else why we are setting up web hooks for the given event

PR #258, opened 2020-01-20
add(docs): detailed information about configuration options

chdsbd:
CHANGES
- break out config information into HTML. This will allow algolia to index our config options.
- update the CSS for the right sidebar to hide overflowing text with an ellipsis instead of allowing wrapping.
- update large configuration example with updating documentation.

### DIFF
--- a/docs/docs/config-reference.md
+++ b/docs/docs/config-reference.md
@@ -126,6 +126,8 @@ This option only applies when `merge.require_automerge_label` is enabled.
 
 Don't wait for in-progress status checks on a PR to finish before updating the branch.
 
+This setting can speed up merges but increases chance of running extra CI jobs.
+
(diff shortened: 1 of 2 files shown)


docs/docs/config-reference.md
```diff
@@ -3,120 +3,414 @@ id: config-reference
...
+automerge label.
+
+This option only applies when `merge.require_automerge_label` is enabled.
+
+### `merge.optimistic_updates`
+
+- **type:** `boolean`
+- **default:** `true`
+
+Don't wait for in-progress status checks on a PR to finish before updating the branch.
```
sbdchd: maybe elaborate on why someone would want this

something like `This setting can speed up merges but increases chance of running CI unnecessarily`

Issue #261, opened 2020-01-25
Kodiak does not update a PR


staabm:
From time to time we experience that kodiak does not update a PR even, if new commits are made in the target branch

E.g. https://github.com/redaxo/redaxo/pull/3210#issuecomment-578396073

Our config
```
# .kodiak.toml
version = 1

[merge]
method = "squash"
delete_branch_on_merge = true
dont_wait_on_status_checks = ["WIP"] # handle github.com/apps/wip
# label to use to enable Kodiak to merge a PR
automerge_label = "automerge" # default: "automerge"
# require that the automerge label be set for Kodiak to merge a PR. if you
# disable this Kodiak will immediately attempt to merge every PR you create
require_automerge_label = true

[merge.message]
title = "pull_request_title"
body = "empty"
include_pr_number = true
strip_html_comments = true # remove html comments to auto remove PR templates

chdsbd:
Thanks for the link, I've checked Sentry and it seems like there are some problems in parsing the webhook payloads for the repository. I've usually ignored the parsing errors but it seems like I was wrong to do so.

I think what happens for your case is that a GitHub webhook event is sent to Kodiak, which Kodiak fails to parse because Kodiak's event schema is incorrect, so Kodiak doesn't evaluate the PR.

I left a review comment https://github.com/redaxo/redaxo/pull/3210#pullrequestreview-348336182 on that linked PR and Kodiak started updating the PR, so I think this is more confirmation that the problem is in the webhook event parsing. I think Kodiak can parse the Pull Request Review webhook event, but fails at some other ones, that's why Kodiak wasn't acting on that PR.

I'm going to simplify the event parsing schemas which I believe will resolve this issue.

I also suspect that this problem may be the cause of https://github.com/chdsbd/kodiak/issues/189.

Thanks for raising this issue.

chdsbd:
I've released #262 which should resolve this issue.

Let me know if you run into any more trouble.

clxmstaab:
we are still seeing some PRs not getting updated.

in a private repo, the PR https://github.com/complex-gmbh/kunzmann/pull/656

![image](https://user-images.githubusercontent.com/47448731/74349702-e6366d00-4db4-11ea-9cc4-6fe345fad962.png)


as you can see on the enabled "Update Branch" there are changes in the target-branch which are not yet merged.. but kodiak does not merge them

chdsbd:
@clxmstaab can you post your configuration file?

[...]

PR #262, opened 2020-01-25
fix: incorrect webhook event schemas

chdsbd:
Our webhook event schemas do not match what GitHub sends us. Instead of trying to match GitHub events 100% we can remove all the fields we don't care about and prevent most of the schema parsing issues.

related #261


kodiak/main.py
```diff
@@ -117,18 +122,17 @@ def find_branch_names_latest(sha: str, branches: List[Branch]) -> List[str]:
...
     """
-    assert status_event.installation
     owner = status_event.repository.owner.login
     repo = status_event.repository.name
     installation_id = str(status_event.installation.id)
     log = logger.bind(owner=owner, repo=repo, install=installation_id)
 
     refs = find_branch_names_latest(
-        sha=status_event.commit.sha, branches=status_event.branches
+        sha=status_event.sha, branches=status_event.branches
```
chdsbd: There's a `sha` at the root of the payload as well as on the commit. To keep the event simple we now only parse that top level `sha`.
https://developer.github.com/v3/activity/events/types/#statusevent

kodiak/main.py
```diff
@@ -83,11 +89,10 @@
 
 
 @webhook()
-async def check_run(check_run_event: events.CheckRunEvent) -> None:
+async def check_run(check_run_event: CheckRunEvent) -> None:
     """
     Trigger evaluation of all PRs included in check run.
     """
-    assert check_run_event.installation
```
chdsbd: we don't need any of these assertions on installation anymore because it's not optional in the schema now.

kodiak/github/test_webhook.py
```diff
@@ -44,7 +44,7 @@ def test_correct_case(
     hook_run = False
 
     @webhook()
-    def push(data: events.PullRequestEvent) -> None:
+    async def push(data: events.PullRequestEvent) -> None:
```
chdsbd: I got rid of the incorrect "sync" support

kodiak/github/fixtures/pull_request_event.json
```diff
@@ -446,5 +446,8 @@
         "received_events_url": "https://api.github.com/users/Codertocat/received_events",
         "type": "User",
         "site_admin": false
+    },
```
chdsbd: all webhook payloads for our GitHub app will have an installation structure.

kodiak/events/check_run.py
```diff
@@ -0,0 +1,32 @@
...
+class CheckRun(pydantic.BaseModel):
+    name: str
+    pull_requests: List[PullRequest]
+
+
+class Owner(pydantic.BaseModel):
+    login: str
+
+
+class Repository(pydantic.BaseModel):
```
chdsbd: The lack of sharing of schema between events is intentional.

kodiak/github/test_webhook.py
```diff
@@ -153,34 +154,27 @@ def test_event_parsing(
...
 
     res = client.post(
         "/api/github/hook",
         data=body,
-        headers={"X-Github-Event": event._event_name, "X-Hub-Signature": sha},
+        headers={"X-Github-Event": event_name, "X-Hub-Signature": sha},
     )
     assert res.status_code == status.HTTP_200_OK
-    assert hook_run == 2, "push and push_async should both be called"
+    assert hook_run == 1, "push_async should both be called"
```
sbdchd: nit
```suggestion
    assert hook_run == 1, "push_async should be called"
```

PR #280, opened 2020-02-04
add(web-api): initial setup for web api

chdsbd:
This PR starts our web api for the web dashboard. This API enables login using GitHub. I'm trying to keep this PR small so the full API build out will be in other PRs.

I originally was going to use FastAPI but gave up after discovery the dearth of libraries. I considered Django but wanted to try something new, so this PR is using Flask with SQLAlchemy.

CHANGES
- copy scripts from `bot/` and modify them to work for this project
- update circleci config to have jobs for this project. I created a new repository on Docker Hub for this `cdignam/kodiak-web-api` and have that set in `s/build`.
- organize circleci jobs into workflows for each project (bot, docs, web_api)
- add sqlalchemy models for database schema to support user login. I had to modify the auto generated migration to enable the `pgcrypto` extension so we can generate UUIDs in postgres. I'm using UUIDs for the primary keys because it seems interesting. I configured server_default with some of the models to make creating data through psql easier.
- add test setup to support sqlalchemy. This was a huge pain and made me long for Django, which has this stuff by default. I found `pytest-flask-sqlalchemy` which seems pretty good for replicating the test transaction behavior of Django. It provides some fixtures that you need to add and then it generally "just works".
- add DIY solution to run migrations before tests. It seems like the common thing to do is to _not_ run the migrations for tests, but that seems silly so there is now some logic to create a test database if it doesn't exist and run the migrations. We reuse the database, so if it gets broken, the best course of action is deleting all the tables so the migrations will set it up again.
- add endpoints for github oauth flow. This isn't super robust but it's a start and should allow us to work on the ui.


chdsbd:
closing in favor of https://github.com/chdsbd/kodiak/pull/281

PR #284, opened 2020-02-10
add: usage reporting via webhook events

chdsbd:
We can calculate usage information purely from `pull_request`, `pull_request_review`, and `pull_request_comment` events.

The GitHub webhook entrypoint is in `bot`, so we need to forward these interesting events to `web_api`. Redis seems like an okay solution since it is already a dependency and allows for the `web_api` to read at it's own pace. We cap the amount of events stored in Redis to prevent Redis from running out of memory if `web_api` stops reading events.

We get about 6-7 million webhook events a month, but I think most of those are `check_run` or `status` events, so I'm guessing pull_request ones are a small fraction of that total.

CHANGES
- update `bot` to place "interesting" webhook events in a queue for `web_api` to read. Compressing the payload reduces our memory usage by 10x (24kB -> 2.4kB).
- add `web_bot` script to pull events from Redis and place them in Postgres for analysis. I created a `github_event` table to store the payload in a jsonb column. We'll add indexes when we figure out what we need to query.


bot/kodiak/event_handlers.py
```diff
@@ -174,7 +178,40 @@ def get_branch_name(raw_ref: str) -> Optional[str]:
             )
 
 
+_redis = None
```
sbdchd: should this be a ContextVar rather than a global?

chdsbd: we share this amongst tasks. Is that invalid?

PR #287, opened 2020-02-11
tools(web_ui): remove compiler error screen from webpack config

sbdchd:
Create React App has an error message screen that pops up when
there is an error from typescript. Sometimes these are type errors, but
a lot of the time they're errors around dead code, unused variables, etc.

Instead we can run typescript in a terminal window or use vscode. Less
intrusive to development.

Also needed to fudge with the start script provided by create react app
so that it wouldn't hang at a message about waiting for type check
results.

Removed some other unnecessary webpack plugins.

### DIFF
--- a/web_ui/config/webpack.config.js
+++ b/web_ui/config/webpack.config.js
@@ -1,7 +1,5 @@
-const fs = require("fs")
 const path = require("path")
 const webpack = require("webpack")
-const resolve = require("resolve")
 // @ts-ignore
 const PnpWebpackPlugin = require("pnp-webpack-plugin")
 const HtmlWebpackPlugin = require("html-webpack-plugin")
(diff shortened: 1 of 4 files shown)


web_ui/config/webpack.config.js
```diff
@@ -345,25 +329,6 @@ module.exports = function(webpackEnv) {
       rules: [
         // Disable require.ensure as it's not a standard language feature.
         { parser: { requireEnsure: false } },
-
-        // First, run the linter.
-        // It's important to do this before Babel processes the JS.
-        {
-          test: /\.(js|mjs|jsx|ts|tsx)$/,
```
sbdchd: we run these in CI or in a terminal session

PR #289, opened 2020-02-12
add(web): mock api

chdsbd:
This change adds mock endpoints that return static data. The web api and web ui are still very much work-in-progress.


web_ui/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.formatOnSave": true
```
sbdchd: I'm thinking we have it so you open vscode in the root of the repo

thoughts?

web_ui/s/build
```diff
@@ -0,0 +1 @@
+exec node scripts/build.js
```
sbdchd: might want to put some shebangs on these scripts

web_ui/package.json
```diff
@@ -93,11 +93,6 @@
     "webpack": "4.41.2",
     "webpack-dev-server": "3.9.0"
   },
-  "scripts": {
-    "start": "node scripts/start.js",
```
sbdchd: should update the web_ui readme since they reference these

web_ui/.vscode/settings.json
chdsbd: I'm doubtful that will work with two python projects, no?

PR #292, opened 2020-02-13
wip: add(bot): better error reporting and context

chdsbd:
All of our logs should have installation ids associated them so we can better connect logs to installations. Sentry should also be configured to do the same. Right now the sentry scoping we have doesn't work because it's not async safe.

CHANGES
- Fix structlog sentry integration to index install, repo, owner. This should allow us to search errors logged through console in Sentry.
- Add context to logs to include installation ids.
- Configure Sentry to have correct context

TODO
- can we use `before_send` to prevent duplicating tags between structlog and sentry? https://docs.sentry.io/error-reporting/configuration/filtering/?platform=python#before-send

docs: http://www.structlog.org/en/stable/contextvars.html

related #261


chdsbd:
I don't think we need to salvage this PR.

PR #293, opened 2020-02-15
add(web): installation models and account endpoints

chdsbd:
- Add Installation model to match the GitHub installation API resource
- Add logic to sync installations and installation memberships for a user
- Migrate to use our own internal IDs instead of relying on GitHub's
- Build out `current_account` and `accounts` endpoints
- Add `sync_accounts` endpoint to trigger syncing of installations
- Update account chooser to handle zero installations and syncing of installations
- Remove user profile and name from sidebar because it's confusing when the user is looking at their Account.

TODO in future PR
- add a job queue to run tasks like syncing installations outside of the request/response cycle.
- handle webhooks for installation events and user auth/deauth events.


web_api/core/models.py
```diff
@@ -38,3 +117,28 @@ class GitHubEvent(BaseModel):
 
     class Meta:
         db_table = "github_event"
+
+
+class Account(BaseModel):
```
sbdchd: doc string explaining the difference between `User`

web_api/core/models.py
```diff
@@ -25,6 +34,76 @@ class Meta:
...
         return True
 
+    def sync_accounts(self) -> None:
+        """
+
+        - create any missing installations
+        - add memberships of user for installations
+        - remove memberships of installations that aren't included
+        """
+        user_installations_res = requests.get(
```
sbdchd: does requests have a default timeout?

web_api/core/models.py
```diff
@@ -38,3 +117,28 @@ class GitHubEvent(BaseModel):
...
+
+
+class Account(BaseModel):
+    class AccountType(models.TextChoices):
+        user = "User"
+        organization = "Organization"
+
+    github_installation_id = models.IntegerField(unique=True)
+    github_account_id = models.IntegerField(unique=True)
+    github_account_login = models.CharField(unique=True, max_length=255)
```
sbdchd: this could use a `help_text`

web_api/core/models.py
```diff
@@ -38,3 +117,28 @@ class GitHubEvent(BaseModel):
...
+    github_account_type = models.CharField(max_length=255, choices=AccountType.choices)
+    payload = pg_fields.JSONField(default=dict)
```
sbdchd: this could also use a `help_text`

web_api/core/models.py
```diff
@@ -38,3 +117,28 @@ class GitHubEvent(BaseModel):
...
+
+    class Meta:
+        db_table = "account"
+
+
+class AccountMembership(BaseModel):
```
sbdchd: doc string

web_api/core/models.py
chdsbd: no, I've configured the timeout parameter, but I don't think it actually sets a hard timeout, just a timeout on how long to wait for more of a response. I think setting a timeout in gunicorn will be the best way to do this.

https://docs.gunicorn.org/en/stable/settings.html#timeout

web_api/core/views.py
```diff
@@ -145,30 +125,11 @@ def accounts(request: HttpRequest) -> HttpResponse:
...
-                profileImgUrl="https://avatars0.githubusercontent.com/u/33015070?s=200&v=4",
-            ),
-            dict(
-                id=8897583,
-                name="pytest-dev",
-                profileImgUrl="https://avatars1.githubusercontent.com/u/8897583?s=200&v=4",
-            ),
+                id=x.id,
+                name=x.github_account_login,
+                profileImgUrl=f"https://avatars1.githubusercontent.com/u/{x.github_account_id}?s=400&v=4",
```
sbdchd: seems like this could be a method on the User model

PR #298, opened 2020-02-25
add(web): management commands for aggregations and docker for web_ui

chdsbd:
To have a staging environment we cannot use Netlify for the react app. This change adds a docker container to serve our React app instead.
I've also moved all of the aggregations into django management commands because I was having we're pathing issues on our servers and this was a clean fix.

CHANGES
- move web_ui from Netlify to docker container
- move aggregations into django managements commands
- handle database conflicts for user_pull_request_activity
- add graceful termination for event ingestion
- configure Sentry for backend and add endpoint to verify Sentry is working
- configure logging for django
- add logic to determine web api domain from app hostname

This code is the currently deployed app.


web_api/Dockerfile
```diff
@@ -15,10 +15,4 @@ RUN poetry install
...
 COPY . /var/app
 
-# workaround for: https://github.com/sdispater/poetry/issues/1123
-RUN rm -rf /var/app/pip-wheel-metadata/
-
-# install cli
-RUN poetry install
-
-CMD poetry run gunicorn --bind 0.0.0.0:$PORT web_api.wsgi
+CMD /var/app/.venv/bin/gunicorn --bind 0.0.0.0:5000 web_api.wsgi
```
sbdchd: I think this should use the array form instead of the string form so signals get sent

web_api/core/management/commands/aggregate_user_pull_request_activity.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Command(BaseCommand):
+    help = "Generate User pull request activity analytics"
+
+    def handle(self, *args, **options):
+        UserPullRequestActivity.generate()
+
+
+# TODO: Move all teh crons to management commands to pathing works prorperly!
```
sbdchd: Still a todo?

web_api/core/test_user_pull_request_activity.py
```diff
@@ -82,6 +82,31 @@ def test_generate(
     assert UserPullRequestActivityProgress.objects.count() == 1
 
 
+@pytest.mark.django_db
+def test_generate_with_conflict(
+    pull_request_total_opened: object,
+    pull_request_kodiak_updated: object,
+    pull_request_kodiak_updated_different_institution: object,
+    pull_request_total_opened_different_institution: object,
+) -> None:
```
sbdchd: Doc comment

web_api/core/views.py
```diff
@@ -320,3 +320,7 @@ def sync_accounts(request: HttpRequest) -> HttpResponse:
     except SyncAccountsError:
         return JsonResponse(dict(ok=False))
     return JsonResponse(dict(ok=True))
+
+
+def debug_sentry(request: HttpRequest) -> HttpResponse:
```
sbdchd: Is this still necessary?

chdsbd: I think it's nice to have a quick way to verify that sentry is working.

.circleci/config.yml
```diff
@@ -290,14 +302,10 @@ workflows:
...
     jobs:
       - web_api_test
       - web_api_lint
-      - web_api_build_container:
-          filters:
-            branches:
-              only: master
-            tags:
-              only: /^v.*/
+      - web_api_build_container
```
chdsbd: It's nice to always build these so we can deploy PRs

PR #305, opened 2020-03-06
add(web_ui): redirect for authed url

chdsbd:
If a user attempts to access an authenticated page but is not logged in, they are redirected to the login page and are redirected to the root page upon successful authentication.

This change will result in the user being redirected to the page they attempted to access but were missing authentication. For example, if a user attempts to access https://app.kodiakhq.com/t/f8c69022-fcba-42a5-9e85-da3a466e6008/ without authentication, after they login successfully, the will be redirected back to that page!

CHANGES
- add login redirects
- update dev script to use new domain

### DIFF
--- a/web_ui/src/auth.ts
+++ b/web_ui/src/auth.ts
@@ -3,7 +3,9 @@ import uuid from "uuid/v4"
 
 export function startLogin() {
   const url = new URL(loginUrl)
-  const state = uuid()
+  const queryParams = new URLSearchParams(location.search)
+  const redirectUri = queryParams.get("redirect")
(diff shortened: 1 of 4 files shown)


web_ui/src/auth.ts
```diff
@@ -13,3 +15,15 @@ export function startLogin() {
 export function getOauthState() {
   return localStorage.getItem("oauth_state") || ""
 }
+
+export function getRedirectPath(x: string): string | undefined {
+  try {
+    const redirect = JSON.parse(x)["redirect"]
```
sbdchd: why json.parse?

PR #307, opened 2020-03-07
fix(web_ui): page loading states having scroll bars

sbdchd:
Our mix of `h-100` on the containers and their parents results in scroll
bars on the loading states for the pages.

Some css changes to remove the scroll bars and center the loading spinner
in the center of the page.

### DIFF
--- a/web_ui/src/components/ActivityPage.tsx
+++ b/web_ui/src/components/ActivityPage.tsx
@@ -1,5 +1,4 @@
 import React from "react"
-import { Container } from "react-bootstrap"
 import { WebData } from "../webdata"
 import { Spinner } from "./Spinner"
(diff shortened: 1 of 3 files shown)


web_ui/src/components/ActivityPage.tsx
```diff
@@ -63,16 +63,12 @@ function ActivityPageInner({ data }: IActivityPageInnerProps) {
...
-  readonly className?: string
 }
-function ActivityPageContainer({
-  children,
-  className,
-}: IActivityPageContainer) {
+function ActivityPageContainer({ children }: IActivityPageContainer) {
   return (
-    <Container className={className}>
+    <div className="flex-grow-1 d-flex flex-column">
```
sbdchd: the container here was adding extra padding as well, not necessary since the page is already wrapped in a `container`

PR #309, opened 2020-03-08
add(web_api): GitHub organization membership roles

chdsbd:
Add api calls to retrieve and store the user's roles for their organizations. This will allow us to restrict billing access to GitHub organization admins only.

Making all these API calls in the request/response cycle is inefficient, but I'm okay with that for the time savings.


web_api/core/models.py
```diff
@@ -209,15 +228,24 @@ class AccountMembership(BaseModel):
...
+        admin = "admin"
+        member = "member"
+
     account = models.ForeignKey(
         Account, on_delete=models.CASCADE, related_name="memberships"
     )
     user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="memberships")
+    role = models.CharField(
+        max_length=255,
+        choices=AccountRole.choices,
```
sbdchd: check constraint for these choices

see: https://adamj.eu/tech/2020/01/22/djangos-field-choices-dont-constrain-your-data/

PR #317, opened 2020-03-10
update(dashboard): sidebar to better support mobile devices

sbdchd:
Simple stacking solution to move the nav to the top of the page when
on a mobile device.

Means we can actually use the site on a phone


web_ui/src/custom.scss
```diff
@@ -58,16 +58,11 @@ a:hover {
   box-shadow: inset 0 0.15em 0.3em rgba(27, 31, 35, 0.15);
 }
 
-.overflow-auto {
-  overflow: auto !important;
```
sbdchd: already covered by bootstrap

PR #347, opened 2020-04-27
add: channel for triggering bot to reevaluate PRs

chdsbd:
When a user hits the paywall, they will need to start a trial or subscription to remove it. However, Kodiak only evaluates a PR when it gets a new related GitHub webhook. So after starting a trial or subscription the user would still see paywalls on their PRs unless they trigger a refresh by updating the PR.

This change adds a method for triggering the bot to reevaluate PRs outside of GitHub webhook triggers. This change adds a process that will listen for events in Redis and trigger updates in the bot.

On trial or subscription upgrades the web api will send a refresh message via Redis to the listening process. The listening process will enqueue webhook events for all open PRs into the webhook event queue, like a normal GitHub webhook event, triggering reevaluation of PRs.

I've done manual testing to verify this works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,12 +306,7 @@ workflows:
     jobs:
       - bot_test
       - bot_lint
-      - bot_build_container:
-          filters:
-            branches:
-              only: master
-            tags:
-              only: /^v.*/
+      - bot_build_container
   docs:
     jobs:
       - docs_typecheck
--- a/bot/kodiak/cli.py
+++ b/bot/kodiak/cli.py
@@ -78,7 +78,9 @@ def validate_config(config_path: str) -> None:
     assert isinstance(cfg_file, V1)
     click.echo(cfg_file.json(indent=2))
 
+
 @cli.command()
 def refresh_pull_requests() -> None:
     from kodiak.refresh_pull_requests import main
+
     main()
--- a/bot/kodiak/cli.py
+++ b/bot/kodiak/cli.py
@@ -79,8 +79,11 @@ def validate_config(config_path: str) -> None:
     click.echo(cfg_file.json(indent=2))
 
 
-@cli.command()
+@cli.command(help="listen for messages and trigger pull request refreshes")
 def refresh_pull_requests() -> None:
+    """
+    Listen on a Redis list for messages triggering pull request reevaluations.
+    """
     from kodiak.refresh_pull_requests import main
 
     main()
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -306,12 +306,7 @@ workflows:
     jobs:
       - bot_test
       - bot_lint
-      - bot_build_container:
-          filters:
-            branches:
-              only: master
-            tags:
-              only: /^v.*/
+      - bot_build_container
```
sbdchd, line 309: should this change be here?

bot/kodiak/cli.py
```diff
@@ -77,3 +77,10 @@ def validate_config(config_path: str) -> None:
     cfg_file = V1.parse_toml(cfg_text)
     assert isinstance(cfg_file, V1)
     click.echo(cfg_file.json(indent=2))
+
+
+@cli.command()
+def refresh_pull_requests() -> None:
```
sbdchd, line 83: doc comment

bot/kodiak/refresh_pull_requests.py
```diff
@@ -0,0 +1,199 @@
...
+logging.basicConfig(
+    stream=sys.stdout,
+    level=conf.LOGGING_LEVEL,
+    format="%(levelname)s %(name)s:%(filename)s:%(lineno)d %(message)s",
+)
+
+# disable sentry logging middleware as the structlog processor provides more
+# info via the extra data field
+# TODO(sbdchd): waiting on https://github.com/getsentry/sentry-python/pull/444
+# to be merged & released to remove `# type: ignore`
```
sbdchd, line 34: comment is old, but that's probably better suited for a different PR

bot/kodiak/refresh_pull_requests.py
```diff
@@ -0,0 +1,199 @@
+from __future__ import annotations
```
sbdchd, line 1: could add a module level doc comment

bot/kodiak/refresh_pull_requests.py
```diff
@@ -0,0 +1,199 @@
...
+        pr_refresh_message = RefreshPullRequestsMessage.parse_raw(res.value)
+        installation_id = pr_refresh_message.installation_id
+        start = time.time()
+        await refresh_pull_requests_for_installation(
+            installation_id=installation_id, redis=redis
+        )
+        logger.info(
+            "pull_request_refresh",
+            installation_id=installation_id,
+            duration=time.time() - start,
```
sbdchd, line 194:
```suggestion
            duration=time.monotonic() - start,
```

.circleci/config.yml
chdsbd: yeah, I don't want it anymore

bot/kodiak/refresh_pull_requests.py
chdsbd: Monotonic isn't what you think it is. It's like a program counter, it starts from the start of the program.

```
❯ python3 -c 'import time; print(time.monotonic())'
0.039981657
```

PR #355, opened 2020-05-01
add: missing subscription in Redis should not block access

chdsbd:
If an account is missing subscription information in Redis, don't block access. Subscription blocking information will be set via the web_api.

This change is what we had originally. I think this is a bit nicer for new users because we don't need to immediately require them to start a trial or subscription. It also means we can elect to ignore hitting certain users with paywalls from the web_api side.

### DIFF
--- a/web_api/core/test_account.py
+++ b/web_api/core/test_account.py
@@ -160,7 +160,7 @@ def test_get_subscription_blocker_seats_exceeded_no_sub_or_trial_no_activity(
     mocker: Any,
 ) -> None:
     """
-    If an account has no trial or subscription, but also not active users, we
+    If an account has no trial or subscription, but also no active users, we
     should not raise the paywall.
     """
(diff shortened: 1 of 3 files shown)


web_api/core/test_account.py
```diff
@@ -155,6 +155,29 @@ def test_get_subscription_blocker_seats_exceeded_no_sub_or_trial(mocker: Any) ->
     assert account.get_active_user_count() == 5
 
 
+@pytest.mark.django_db
+def test_get_subscription_blocker_seats_exceeded_no_sub_or_trial_no_activity(
+    mocker: Any,
+) -> None:
+    """
+    If an account has no trial or subscription, but also not active users, we
```
chdsbd, line 163:
```suggestion
    If an account has no trial or subscription, but also no active users, we
```

PR #365, opened 2020-05-08
add(web_ui): billing history link via stripe self serve

sbdchd:
Add a link to Stripe's self serve portal for viewing billing history.

https://stripe.com/docs/billing/subscriptions/integrating-self-serve-portal

We cannot use the self-serve portal for updating subscriptions because the portal does not support editing subscriptions with quantities (like Kodiak seat licenses). We could use it for editing payment methods, but we already have a way that does that nicely through Stripe.

This PR updates the stripe sdk to a version that supports the self-serve portal.

Screenshot 2020-05-07 19 55 18


web_api/core/test_stripe_views.py
```diff
@@ -0,0 +1,51 @@
+import pytest
```
chdsbd, line 1: You want to open another PR to move break up the tests in test_views.py? 😄

PR #368, opened 2020-05-09
add(web): make Kodiak free for personal accounts

chdsbd:
This change makes Kodiak free for personal GitHub accounts. Kodiak now only requires a subscription for private GitHub organization repositories.

The subscription section is replaced with some copy saying Kodiak is free for personal accounts.
image

### DIFF
--- a/web_ui/src/api.ts
+++ b/web_ui/src/api.ts
@@ -33,6 +33,7 @@ export interface IUsageBillingPageArgs {
   readonly teamId: string
 }
 export interface IUsageBillingPageApiResponse {
+  readonly canSubscribe: boolean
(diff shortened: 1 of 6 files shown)


web_ui/src/api.ts
```diff
@@ -33,6 +33,7 @@ export interface IUsageBillingPageArgs {
   readonly teamId: string
 }
 export interface IUsageBillingPageApiResponse {
+  readonly canSubscribe: boolean
```
sbdchd, line 36: has the api been updated to return this?

chdsbd: fixed

Issue #276, opened 2020-01-31
leave a comment when PR contains conflicts


clxmstaab:
we have some long running PRs which kodiak permantly rebases to stay up 2 date with the target branch.

sometimes those PR get stuck because the PR gets into a conflict state.
we dont notice it though that the PR no longer updates because of conflicts.

it would be great if kodiak could leave a comments within the PR so people working on it are aware that the PR cannot be updated automatically anymore because of conflicts

chdsbd:
Do you know if you’re encountering a normal merge conflict or a rebase merge conflict?
There’s [logic to notify about normal conflicts](https://kodiakhq.com/docs/config-reference#mergenotify_on_conflict) but not rebase ones.

Also, when these PRs don’t update is there an error status message from Kodiak?

This may be related: https://github.com/chdsbd/kodiak/issues/222

clxmstaab:
how to differentiate a "merge conflict" from a "rebase merge conflict"?

I guess it might be the "rebase merge conflict" case, because its related to a PR which is blocked by a "WIP:" title

chdsbd:
I think I figured out the issue and this isn't related to rebase conflicts as I first thought.

If you're using `update.always`, you won't get notified about merge conflicts that occur when updating because Kodiak returns before it hits that logic:

https://github.com/chdsbd/kodiak/blob/9d98de1d694d21053abd6af81792cc6f04aa622a/kodiak/evaluation.py#L300-L311

https://github.com/chdsbd/kodiak/blob/9d98de1d694d21053abd6af81792cc6f04aa622a/kodiak/evaluation.py#L373-L388

I think the `notify_on_conflict` config option should be a top level field instead of being nested in the `merge` options because I think a user should always be notified of conflicts.

So we need a new option `notify_on_conflict` that will notify whenever there is a conflict and deprecate `merge.notify_on_conflict`.

I think this new option should resolve your issue.

example:

```toml
# old
[merge]
notify_on_conflict = true

[update]
always = true
```

```toml
# new
notify_on_conflict = true

[update]
always = true
```

staabm:
> I think the notify_on_conflict config option should be a top level field instead of being nested in the merge options because I think a user should always be notified of conflicts.
So we need a new option notify_on_conflict that will notify whenever there is a conflict and deprecate merge.notify_on_conflict.

Sounds reasonable

clxmstaab:
[...]

PR #364, opened 2020-05-08
add: notify_on_conflict to notify on any merge conflict

chdsbd:
This change adds two configuration options: `notify_on_conflict` and `notify_on_conflict_label`. If a PR encounters a merge conflict, Kodiak will add a label (configurable via `notify_on_conflict_label`) and leave a comment, even if the PR isn't set for merging. When the merge conflict is resolved, Kodiak will remove the label.

We have an existing `merge.notify_on_conflict` option, but Kodiak will only notify when the automerge label is configured. `notify_on_conflict` also supersedes `merge.notify_on_conflict` (the code is earlier in evaluation.py).

Here's an example config that enables `notify_on_conflict` and sets a custom `notify_on_conflict_label`.

```toml
# .kodiak.toml
version = 1

notify_on_conflict = true
notify_on_conflict_label = "merge conflict!?"

[merge]
method = "squash"
```

While I think `notify_on_conflict` would be a good default option, I think we should leave it opt-in to preserve existing behavior for users. (edit: I'm not sure about this)

I'll document this feature in another PR after it's been deployed.

fixes #276

PR #371, opened 2020-05-10
add(bot): prioritize merge conflict notifications

chdsbd:
fixes https://github.com/chdsbd/kodiak/issues/276

This problem is #276 is that the `merge.blacklist_title_regex` was taking priority over `merge.notify_on_conflict`. This is not as desirable because I think merge conflict notifications are higher priority.

I was going to make a more general change in #364, but I think this is a simple solution that addresses the immediate problem. I plan to circle back to #364 later.

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -381,6 +381,28 @@ async def mergeable(
             f"missing automerge_label: {config.merge.automerge_label!r}",
         )
         return
+
+
+    # We want users to get notified a merge conflict even if the PR matches a
+    # WIP title via merge.blacklist_title_regex.
+    if (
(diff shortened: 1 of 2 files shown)


bot/kodiak/evaluation.py
```diff
@@ -381,6 +381,28 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
             f"missing automerge_label: {config.merge.automerge_label!r}",
         )
         return
+
+
+    # We want users to get notified a merge conflict even if the PR matches a
```
chdsbd, line 386: This change is moving the merge.notify_on_conflict logic above other checks.

PR #375, opened 2020-05-11
fix(web_api): oauth flow not handling api error in membership check

chdsbd:
If the user is a collaborator of an organization instead of a member or admin, calling GET https://api.github.com/orgs/:org/memberships/:user will fail with a permission error. This change handles that case.


chdsbd:
This change has been deployed.

PR #381, opened 2020-05-16
fix(bot): merge.update_branch_immediately breaking out of merge loop

chdsbd:
I realized we were breaking out of the merge loop for this, which isn't desirable behavior. Now we correctly continue polling while merging.

depends on #380


bot/kodiak/evaluation.py
```diff
@@ -680,6 +680,8 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
                 markdown_content="branch updated because `merge.update_branch_immediately = true` is configured.",
             )
             await api.update_branch()
+            if merging:
```
sbdchd, line 683: is the comment above this still needed?

chdsbd: fixed

PR #383, opened 2020-05-16
ref(bot): simplify evaluation testing to be more readable

chdsbd:
We have a lot of unnecessary information in these evaluation.py tests. This change makes it so we only have information that's important to the test.

We're basically replacing all the pytest fixtures with plain functions.

I've changed one test as an example. I plan to make more changes in chunks via other PRs.

### DIFF
--- a/bot/kodiak/test_evaluation.py
+++ b/bot/kodiak/test_evaluation.py
@@ -5,6 +5,7 @@ from typing import Any, List, Mapping, Optional, Tuple, Union
 import pydantic
 import pytest
 from toml import TomlDecodeError
+from typing_extensions import Protocol
 
 from kodiak.config import (
     V1,
@@ -379,7 +380,7 @@ def create_repo_info() -> RepoInfo:
     )
 
 
-def create_api() -> PRAPI:
+def create_api() -> MockPrApi:
     return MockPrApi()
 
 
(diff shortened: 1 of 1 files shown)


bot/kodiak/test_evaluation.py
```diff
@@ -355,45 +534,21 @@ async def mergeable(
...
-        api_call_retry_method_name=None,
-        #
-        is_active_merge=True,
-    )
+    await mergeable(api=api, is_active_merge=True)
     assert api.queue_for_merge.called is True
 
+    assert (
+        api.set_status.call_count == 0
+    ), "we don't want to set a status message from the frontend when the PR is being merged"
```
chdsbd, line 551: I think this is something that should have been a part of the test originally.

bot/kodiak/test_evaluation.py
```diff
@@ -280,25 +332,152 @@ def review() -> PRReview:
...
+            api_call_retry_timeout=api_call_retry_timeout,
+            api_call_retry_method_name=api_call_retry_method_name,
+            subscription=subscription,
+            app_id=app_id,
+        )
+
+    return mergeable
+
+
 async def mergeable(
```
chdsbd, line 481: I'll delete this function once everything is moved over to `create_mergeable`.

bot/kodiak/test_evaluation.py
```diff
@@ -280,25 +332,152 @@ def review() -> PRReview:
...
+    cfg.merge.method = MergeMethod.squash
+    return cfg
+
+
+class MergeableType(Protocol):
+    """
+    A type we define so our create_mergeable() can be typed.
+    """
+
+    async def __call__(
```
chdsbd, line 400: I learned that this is the way to define the function type. `Callable` doesn't have enough detail.
https://mypy-lang.blogspot.com/2018/09/mypy-0630-released.html

bot/kodiak/test_evaluation.py
```diff
@@ -257,6 +273,29 @@ def pull_request() -> PullRequest:
     )
 
 
+def create_pull_request() -> PullRequest:
```
chdsbd, line 276: All of the pytest fixtures will be removed once everything has been migrated to the `create_`-prefixed functions.

PR #398, opened 2020-05-19
fix(bot): disable Kodiak on GitHub API merge internal server error

### LINKED ISSUE #397: It's unsafe to retry merge when GitHub's pull request merge API returns a 500 error
Kodiak has merged the same PR multiple times. It wasn't until I manually clicked the merge button that it stopped.

https://github.com/zeit/now/commits/master

https://github.com/zeit/now/pull/4412

image

chdsbd:
This change disables Kodiak merging a pull request when it encounters an internal server error from the GitHub merge api. In this event, Kodiak adds a label `disable_kodiak_label` and leaves a comment describing the issue. Kodiak will not update, approve, comment on, label, or merge a pull request with the `disable_kodiak_label`. It will still update status checks. The user must remove this label to re-enable Kodiak merging the pull request.

fixes #397

## background
It is dangerous to retry merging when GitHub returns an internal server error (HTTP status code 500) because the merge can partially succeed in a way that will allow Kodiak to erroneously merge in a pull request multiple times, as we've seen in #397.

The GitHub pull request merge API is _not always_ idempotent (https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button). Normally if you try to merge an already merged pull request using this API you'll get a `405 Method not allowed` error, as documented in the GitHub API docs. However in an error state, when the GitHub API returns a 500 HTTP error, the pull request can be partially merged, as shown in #397.

It appears the git operations of merging the pull request branch into master can succeed but not affect the pull request state itself. In this failure scenario the pull request will still be in an OPEN state even though the underlying branch was merged successfully into the target branch.

## further discussion
I realized implementing this change that GitHub API calls to add pull request labels could also fail. In this case Kodiak would still erroneously merge a pull request.

I think the most robust solution to this GitHub API reliability problem would be to store some information in Redis to disable Kodiak. But then there's the problem of how to surface that data in the GitHub UI. If we can't trust the GitHub API, how can we notify a user that Kodiak is disabled on their pull request? 🙍‍♂️ 

## screenshots
image

### DIFF
--- a/bot/kodiak/config.py
+++ b/bot/kodiak/config.py
@@ -124,7 +124,7 @@ class V1(BaseModel):
     merge: Merge = Merge()
     update: Update = Update()
     approve: Approve = Approve()
-    # label to apply to pull request to disable Kodiak from taking any action
+    # when added to a Pull Request Kodiak will be prevented from taking any action
     # (approvals, updates, merges, comments, labels). Kodiak will still set
     # status checks. A user should generally not need to change this label as it
     # should rarely be applied.
--- a/bot/kodiak/config.py
+++ b/bot/kodiak/config.py
@@ -83,10 +83,6 @@ class Merge(BaseModel):
     #
     # immediately update a PR whenever the target updates
     update_branch_immediately: bool = False
-    # label to apply to pull request when Kodiak encounters a 500 error merging
-    # a pull request. This is only used when merge.require_automerge_label is
-    # disabled.
-    merge_failure_label: str = "kodiak:merge-failed"
     # if a PR is passing all checks and is able to be merged, merge it without
     # placing it in the queue. This will introduce some unfairness where those
     # waiting in the queue the longest will not be served first.
@@ -128,6 +124,11 @@ class V1(BaseModel):
     merge: Merge = Merge()
     update: Update = Update()
     approve: Approve = Approve()
+    # label to apply to pull request to disable Kodiak from taking any action
+    # (approvals, updates, merges, comments, labels). Kodiak will still set
+    # status checks. A user should generally not need to change this label as it
+    # should rarely be applied.
+    disable_bot_label: str = "kodiak:disabled"
 
     @validator("version", pre=True, always=True)
     def correct_version(cls, v: int) -> int:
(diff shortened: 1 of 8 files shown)


bot/kodiak/evaluation.py
```diff
@@ -750,11 +757,44 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
...
+        try:
+            await api.merge(
+                merge_method=merge_args.merge_method,
+                commit_title=merge_args.commit_title,
+                commit_message=merge_args.commit_message,
+            )
+        # if we encounter an internal server error (status code 500), it is
+        # _not_ safe to retry. Instead we mark the pull request as unmergable
+        # and require a user to re-enable Kodiak on the pull request.
+        except GitHubApiInternalServerError:
```
chdsbd, line 769: I know we have some retry logic for API exceptions that runs outside of `mergeable()`, but I think this is the proper place to handle this error. I think we should also try to move that retry logic in to the function as well.

We can easily provide thorough test coverage of any logic inside of `mergable()`.

bot/kodiak/pull_request.py
```diff
@@ -283,13 +288,30 @@ async def merge(
...
+        self.log.info("add_label", label=label)
+        async with self.client(
+            installation_id=self.install, owner=self.owner, repo=self.repo
+        ) as api_client:
+            res = await api_client.add_label(label, pull_number=self.number)
+            try:
+                res.raise_for_status()
+            except HTTPError:
+                self.log.exception("failed to delete label", label=label, res=res)
+                raise ApiCallException("add label")
```
chdsbd, line 313: I realized while adding this endpoint that if we encounter an exception trying to add a label to a pull request we will not be able to disable the bot.

I'm not sure what to do about this.

bot/kodiak/config.py
```diff
@@ -124,6 +124,11 @@ class V1(BaseModel):
     merge: Merge = Merge()
     update: Update = Update()
     approve: Approve = Approve()
+    # label to apply to pull request to disable Kodiak from taking any action
```
sbdchd, line 127:
```suggestion
    # when added to a Pull Request Kodiak will be prevented from taking any action
```

chdsbd:
This change has been deployed.

PR #410, opened 2020-05-30
ref: update seat usage to assign users to seats

chdsbd:
Previously if an account went over their seat quantity they would be considered blocked. Now only users in excess of an account's subscription quantity will be blocked.

Basically individual users now occupy seats. If they occupy a seat before an account's seats are consumed, they'll be able to use Kodiak even as excess users are blocked.

This behavior is better because it allows organizations to user the seats they purchased even when they have more users than purchased seats.


web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -982,7 +993,9 @@ function UsageBillingPageInner(props: IUsageBillingPageInnerProps) {
                     {u.name}
                   </td>
                   <td>{u.interactions}</td>
+                  <td>{u.firstActiveDate}</td>
```
sbdchd, line 996: when firstActiveDate is undefined what does this show / what should it show?

web_api/core/views.py
```diff
@@ -80,21 +80,38 @@ def usage_billing(request: HttpRequest, team_id: str) -> HttpResponse:
...
+    )
+    # we assign seats to users in order of first active for the last 30 days.
+    allowed_user_ids = set(
+        (
+            user.github_id
+            for user in sorted(active_users, key=lambda x: x.first_active_at)[
+                :subscription_quantity
+            ]
+        )
+    )
```
sbdchd, line 95:
```suggestion
    allowed_user_ids = {
            user.github_id
            for user in sorted(active_users, key=lambda x: x.first_active_at)[
                :subscription_quantity
            ]
    }
```

web_api/core/models.py
```diff
@@ -691,7 +716,9 @@ def get_active_users_in_last_30_days(account: Account) -> List[ActiveUser]:
                 AND b.is_private_repository = TRUE
                 AND a.github_installation_id = %s
             GROUP BY
-                a.github_user_id;
+                a.github_user_id
+            ORDER BY
```
sbdchd, line 720: are there tests for this change?

web_ui/src/components/UsageBillingPage.tsx
chdsbd: Here's what it looks like when those are undefined. I think this is okay.
image

web_api/core/models.py
chdsbd: fixed

web_api/core/views.py
chdsbd: fixed

PR #412, opened 2020-06-04
fix: docker CMD to handle signals

chdsbd:
Using `poetry run` means signals don't get sent correctly so stopping a container is slow. Calling uvicorn directly ensures signals are properly handled.

This is similar to what we do on the web api container.


chdsbd:
This didn't work. Need to use the array style `CMD` argument.

PR #413, opened 2020-06-04
dep: bump mypy and sentry-sdk

chdsbd:
This allows us to remove some `type: ignore` comments thanks to improved type stubs.


chdsbd:
typeshed isn't updated to make this work

PR #422, opened 2020-06-09
add: support for different currencies in dashboard

chdsbd:
Add support for giving a user a different pricing plan. For example, a user may require that we bill in EUR instead of USD. We can create a pricing plan in Stripe in EUR and configure the dashboard to use that for billing instead of the default plan_id.

Add support for display different currencies in web ui.

image

image

TODO in future PR
- display plan-specific pricing in subscription creation form


web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -662,6 +686,10 @@ function ActiveSubscription({
   modifySubscription,
   teamId,
 }: IActiveSubscriptionProps) {
+  const formatter = new Intl.NumberFormat("en-US", {
```
sbdchd, line 689:
```suggestion
  const formatter = new Intl.NumberFormat(undefined, {
```

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -110,8 +113,12 @@ function UsageAndBillingContainer({ children }: { children: React.ReactNode }) {
   )
 }
 
-function formatCents(cents: number): string {
-  return `\$${(cents / 100).toFixed(2)}`
+function formatCents(cents: number, currency: string): string {
```
sbdchd, line 116: why aren't we using this function down below?

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -691,9 +719,11 @@ function ActiveSubscription({
           <b>Cost</b>
         </Col>
         <Col>
-          <span className="mr-4">{formatCents(cost.totalCents)} / month</span>
+          <span className="mr-4">
+            {formatter.format(cost.perSeatCents / 100)} / month
+          </span>
           <span>
-            ({formatCents(cost.perSeatCents)} / seat ⨉ {seats} seats)
+            ({formatter.format(cost.perSeatCents / 100)} / seat ⨉ {seats} seats)
```
sbdchd, line 726: like here?

web_ui/src/components/UsageBillingPage.tsx
chdsbd: fixed

PR #427, opened 2020-06-13
ref(web_ui): redesign subscription components to be simple and clear

chdsbd:
The original design was kind of rough and I think this new design is a little more professional and clear.

signup page before
inactive-before

signup page after
inactive-after

active trial before
trial-active-before

active trial after
trial-active-after

active subscription before
sub-before

active subscription after
sub-after


web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -578,6 +580,56 @@ function ManageSubscriptionModal({
...
+      }>
+      <Card.Header>
+        <h4 className="text-center">{name}</h4>
+      </Card.Header>
+      <Card.Body className="d-flex flex-column">
+        <h1 className="text-center">{cost}</h1>
+        <div className="flex-grow-1 d-flex flex-column">
+          <ul className="flex-grow-1 list-unstyled mt-3 mb-4 text-center">
+            {features.map(x => (
+              <li>{x}</li>
```
sbdchd, line 611: probably doesn't matter since we aren't reordering the elements or anything, but seems like we don't have eslint-react setup

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -578,6 +580,56 @@ function ManageSubscriptionModal({
...
+  features,
+  startButton,
+  highlight,
+}: {
+  className?: string
+  name: React.ReactNode
+  cost: React.ReactNode
+  features: React.ReactNodeArray
+  startButton: React.ReactNode
+  highlight?: boolean
```
sbdchd, line 596:
```suggestion
  readonly className?: string
  readonly name: React.ReactNode
  readonly cost: React.ReactNode
  readonly features: React.ReactNodeArray
  readonly startButton: React.ReactNode
  readonly highlight?: boolean
```

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -578,6 +580,56 @@ function ManageSubscriptionModal({
...
+    </Card>
+  )
+}
+
+const KodiakTooltip = ({
+  children,
+  content,
+}: {
+  children: React.ReactNode
+  content: React.ReactNode
```
sbdchd, line 626:
```suggestion
  readonly children: React.ReactNode
  readonly content: React.ReactNode
```

PR #431, opened 2020-06-14
add(web_api): fields for editing customer information

chdsbd:
This change adds an endpoint for updating customer info (email, name, address). Email is something customers have needed to alter and these other fields allow users to add more information that will be displayed in receipts if provided.

This change also adds support for the `customer.updated` webhook event from Stripe.


web_api/core/views.py
```diff
@@ -498,6 +553,16 @@ def stripe_webhook_handler(request: HttpRequest) -> HttpResponse:
         logger.warning("more action required for payment %s", event)
     elif event.type == "invoice.payment_failed":
         logger.warning("invoice.payment_failed %s", event)
+    elif event.type in "customer.updated":
```
sbdchd, line 556: is this supposed to be `==` or `in`?

web_api/core/test_views.py
```diff
@@ -697,6 +720,183 @@ class FakeCheckoutSession:
     assert res.json()["stripePublishableApiKey"] == settings.STRIPE_PUBLISHABLE_API_KEY
 
 
+def create_pk() -> int:
+    return random.randint(1000, 100000)
```
sbdchd, line 724: Is there a way to avoid this? Seems like it could lead to flakey tests

chdsbd: fixed

PR #432, opened 2020-06-14
add(web_ui): components for editing email, name, and address

chdsbd:
This change builds the UI to talk to the recently added stripe_customer_info endpoint

app localhost kodiakhq com_3000_t_1c78d8f1-189c-470e-9d61-3b46c478e7e0_usage_ (1)


web_ui/src/world.ts
```diff
@@ -147,5 +147,12 @@ export const Current: World = {
           `/v1/t/${args.teamId}/subscription_info`,
         )
         .then(d => d.data),
+    updateStripeCustomerInfo: (args: api.UpdateStripeCustomerInfoArgs) => {
+      const { teamId, ...payload } = args
+      return authRoute.post<unknown>(
+        `/v1/t/${args.teamId}/update_stripe_customer_info`,
+        payload,
```
chdsbd, line 154: we're using JSON instead of formData. We had to move the api to `app.kodiakhq.com` to fix CORS issues with `application/json` Content Type.

web_ui/src/webdata.ts
```diff
@@ -1,4 +1,5 @@
 export type WebData<T> =
+  | { readonly status: "initial" }
```
chdsbd, line 2: This is necessary for when we have api calls we don't want to call immediately, like mutations.

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -945,7 +1176,7 @@ function UsageBillingPageInner(props: IUsageBillingPageInnerProps) {
   function clearQueryString() {
     history.push({ search: "" })
   }
-  if (props.data.status === "loading") {
+  if (props.data.status === "initial" || props.data.status === "loading") {
```
chdsbd, line 1179: We have to update `WebData<T>` sites.

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -833,6 +1050,20 @@ function Subcription({
             />
           </Card.Body>
         </Card>
+        <BillingEmailForm
+          className="mb-4"
+          defaultValue={subscription.billingEmail}
```
chdsbd, line 1055: We don't update the parent data on save. I think this is okay and it makes things easier to implement.

The slight downside is that if you edit the billing email and open the "Modify Subscription" modal the new email will not be reflected.

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -764,6 +972,15 @@ function Subcription({
       readonly currency: string
     }
     readonly billingEmail: string
```
chdsbd, line 974: I would have called this `customerEmail` if I was doing this now ☹️

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -749,6 +753,210 @@ function SubscriptionUpsellPrompt({
...
+                />
+              </Col>
+            </Form.Row>
+          </Form.Group>
+          <Form.Group>
+            <Form.Control
+              type="text"
+              placeholder="Country"
+              value={country}
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
```
chdsbd, line 945: We have to type all of these params for some reason. Using a regular `<input/>` element doesn't require that.

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -513,11 +522,6 @@ function ManageSubscriptionModal({
           <Form.Group>
             <Form.Label>Billing Email </Form.Label>
             <Form.Control type="text" required disabled value={billingEmail} />
-            <Form.Text className="text-muted">
```
chdsbd, line 516: We don't want to direct people to update there billing email this way because it requires updating the credit card too.

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -749,6 +753,210 @@ function SubscriptionUpsellPrompt({
...
+            <Form.Control
+              type="email"
+              required
+              value={billingEmail}
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+                setBillingEmail(e.target.value)
+              }
+            />
+            <Form.Text className="text-muted">
+              Required. Address to send billing receipts.
```
sbdchd, line 800: I think this required should be made bold, so it sticks out a bit more

web_ui/src/components/UsageBillingPage.tsx
```diff
@@ -749,6 +753,210 @@ function SubscriptionUpsellPrompt({
...
+    defaultValue.postalCode ?? "",
+  )
+  const [country, setCountry] = React.useState(defaultValue.country ?? "")
+
+  const [apiState, updateStripeCustomerInfo] = useTeamApiMutation(
+    Current.api.updateStripeCustomerInfo,
+  )
+  function handleSubmit(e: React.ChangeEvent<HTMLFormElement>) {
+    e.preventDefault()
+    updateStripeCustomerInfo({
```
sbdchd, line 877: how does this handle api failure?

Shouldn't it undo the changes when there is an error?

chdsbd: We display error text "save failed" via the button component.

web_ui/src/components/UsageBillingPage.tsx
chdsbd: image

web_ui/src/components/UsageBillingPage.tsx
chdsbd: image

PR #444, opened 2020-06-19
add(bot): alias for blacklist_

### LINKED ISSUE #437: Rename `blacklist_labels` to `denylist_label`
## Description

The config of Kodiak allows to specify various blacklist types. 

## Solution

I would propose to rename any occurance of blacklist to allowlist to avoid oppressive wording. 

These include:

1. `blacklist_labels`
2. `blacklist_title_regex`
3. `blacklist_usernames`

which can be renamed to

1. `denylist_labels`
2. `denylist_title_regex`
3. `denylist_usernames`

I would propose to perform this in a backwards compatible way so the old blacklist configs continue to work while they are not documented anymore and a part of the documentation informs about the possible migration.

chdsbd:
blacklist has confused some users and has negative connotations. We can use more precise language instead.

fixes #437


sbdchd:
Docs should be updated _after_ this is merged & deployed so users don't start using invalid options

bot/kodiak/test_evaluation.py
```diff
@@ -884,7 +884,7 @@ async def test_mergeable_blacklist_title_regex() -> None:
     assert api.set_status.call_count == 1
     assert api.dequeue.call_count == 1
     assert "cannot merge" in api.set_status.calls[0]["msg"]
-    assert "matches blacklist_title_regex" in api.set_status.calls[0]["msg"]
+    assert "matches merge.blacklist_title_regex" in api.set_status.calls[0]["msg"]
```
chdsbd, line 887: I'm okay with this change.

bot/kodiak/config.py
```diff
@@ -45,16 +45,23 @@ class MergeMessage(BaseModel):
     include_pull_request_url: bool = False
 
 
+# this pattern indicates that the user has the field unset.
+UNSET_TITLE_REGEX = ":::|||kodiak|||internal|||reserved|||:::"
```
chdsbd, line 49: This feels hacky, but it seems to cover our bases.

tdeekens:
Can somebody ping me when it's deployed. Then I'd love to test it right away :)

chdsbd:
@tdeekens This change has been deployed to production.

PR #451, opened 2020-06-21
add(web_api): support annual subscriptions

chdsbd:
Update API to support creating and upgrading to annual subscriptions.


web_api/core/views.py
```diff
@@ -419,10 +445,25 @@ def get_subscription_info(request: HttpRequest, team_id: str) -> JsonResponse:
...
 
+def plan_id_from_period(period: Literal["month", "year"]) -> str:
+    if period == "month":
+        return cast(str, settings.STRIPE_PLAN_ID)
+    if period == "year":
+        return cast(str, settings.STRIPE_ANNUAL_PLAN_ID)
+    return None
+
+
+class FetchProrationModal(pydantic.BaseModel):
```
chdsbd, line 456: We really should have been using Pydantic from the start 🤦

web_api/core/views.py
```diff
@@ -229,23 +238,29 @@ def update_subscription(request: HttpRequest, team_id: str) -> HttpResponse:
...
-    invoice = stripe.Invoice.create(
-        customer=stripe_customer_info.customer_id, auto_advance=True
+        proration_date=payload.prorationTimestamp,
     )
-    # we must specify the payment method because our Stripe customers don't have
-    # a default payment method, so the default invoicing will fail.
-    stripe.Invoice.pay(invoice.id, payment_method=subscription.default_payment_method)
+    # we only need to manually created invoices when a user modifies their
+    # subscription within the same billing period.
+    if payload.planPeriod == subscription.plan.interval:
```
chdsbd, line 249: This is super weird. I don't really understand it, but it's the reality. Users get charged immediately when changing between plans but not when upgrading.

web_api/core/views.py
```diff
@@ -211,14 +215,19 @@ def start_trial(request: HttpRequest, team_id: str) -> HttpResponse:
...
 
+class UpdateSubscriptionModel(pydantic.BaseModel):
+    seats: int
+    prorationTimestamp: int
+    planPeriod: Literal["month", "year"] = "month"
+
+
 @auth.login_required
 def update_subscription(request: HttpRequest, team_id: str) -> HttpResponse:
+    payload = UpdateSubscriptionModel.parse_obj(request.POST.dict())
```
chdsbd, line 226: I wonder if there is a better name than `payload`. 🤷‍♂️

web_api/core/test_views.py
```diff
@@ -472,9 +523,114 @@ def test_update_subscription(
...
+            default_payment_method="pm_22dldxf3",
+        ),
+        "fake-key",
+    )
+
+
+@pytest.mark.django_db
+def test_update_subscription_switch_plans(
+    authed_client: Client, user: User, other_user: User, mocker: Any
+) -> None:
```
sbdchd, line 592: doc comment

web_api/core/views.py
sbdchd: I think `payload` is fine 

Some other options:

- `body`
- `request_body`
- `req`
- `data`
- `request_data`

web_api/core/test_views.py
```diff
@@ -472,9 +523,114 @@ def test_update_subscription(
     assert stripe_subscription_retrieve.call_count == 1
     assert update_bot.call_count == 1
     assert stripe_subscription_modify.call_count == 1
+    _args, kwargs = stripe_subscription_modify.call_args
+    assert kwargs["items"][0]["plan"] == settings.STRIPE_PLAN_ID
     assert stripe_invoice_create.call_count == 1
     assert stripe_invoice_pay.call_count == 1
 
+    res = authed_client.post(
```
sbdchd, line 531: comment on what these are testing

PR #454, opened 2020-06-22
update(docs): documentation for blacklist_ option removal

chdsbd:
Replace config options with there new alias.

I was planning to document each option, but realized that could be confusing. Instead we indicate the former names as aliases and add some notes.


chdsbd:
This will go out when #444 is deployed.

chdsbd:
@sbdchd came up with a hacky solution. The positioning isn't great, but it works.

tdeekens:
Thanks all!

Issue #387, opened 2020-05-16
Use whatever PR merge method is allowed on repo if there's only one


gnarea:
First of all, thanks for building kodiak and making it available!

I configure all my repos to only allow squash merges and I wish I could just use the minimal config file with just `version = 1`, and kodiak detected that it should use squash commits.

However, it currently fails because the merge method is the default.

chdsbd:
Great idea. This should be any easy change.

chdsbd:
I'll update this issue when #464 has been deployed.

chdsbd:
#464 has been deployed to production. If `merge.method` is not configured, Kodiak will use the first valid merge method in the repository based on the list "merge", "squash", "rebase".

gnarea:
Thank you so much!

PR #464, opened 2020-07-06
add(bot): use valid repo merge method if not specified

chdsbd:
If `merge.method` is not configured, use the first valid merge method in the repository based on the list "merge", "squash", "rebase".

For example, if a repository only allows "rebase" commits and `merge.method` is not provided, Kodiak will now use the "rebase" merge method. Previously it would have used the default "merge" method.

If a repository only allowed "squash" and "rebase", and `merge.method` was not configured, "squash" would be chosen because it is the first valid option in the list "merge", "squash", "rebase".

This change is backwards compatible for the most part. Kodiak now works instead of raising a configuration error if a user only allowed "squash" merge methods and didn't specify `merge.method`. If a user was relying on `"merge"` being the default, they will be okay since `"merge"` is the first option in the list of valid merge methods to try.

Fixes https://github.com/chdsbd/kodiak/issues/387

Docs will be updated in https://github.com/chdsbd/kodiak/pull/466.

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -435,7 +435,13 @@ async def mergeable(
         )
         return
 
-    merge_method = config.merge.method
+    if config.merge.method is not None:
+        merge_method = config.merge.method
+    elif len(valid_merge_methods) == 1:
+        merge_method = valid_merge_methods[0]
+    else:
+        merge_method = MergeMethod.merge
+
     if (
         branch_protection.requiresCommitSignatures
         and merge_method == MergeMethod.rebase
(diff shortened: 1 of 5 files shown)


bot/kodiak/evaluation.py
```diff
@@ -435,7 +435,13 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
         )
         return
 
-    merge_method = config.merge.method
+    if config.merge.method is not None:
+        merge_method = config.merge.method
+    elif len(valid_merge_methods) == 1:
+        merge_method = valid_merge_methods[0]
+    else:
+        merge_method = MergeMethod.merge
```
sbdchd, line 443: this seems a little weird. If there is only one then taking the first makes sense, but here if there are multiple we don't choose any of them.

Instead we should pick the first one and maybe log something to sentry if there are multiple values

chdsbd:
This has been deployed.

Issue #401, opened 2020-05-22
self-hosted, kodiakhq: status ⛴ attempting to merge PR (merging) 


haghabozorgi:
We are running kodiak in our kube cluster,  and when we test a test PR, it works as expected, it requires  the PR is merged, which has and the branch is deleted as expected, but the checks tab in github for that merged pr shows

```
kodiakhq: status ⛴ attempting to merge PR (merging) 
```

github checks tab snippet

![Screenshot_20200521_173058](https://user-images.githubusercontent.com/1186748/82618783-df27b380-9b88-11ea-8206-1d20440cfbbd.png)

github repo 'branch' settings

![Screenshot_20200521_173601](https://user-images.githubusercontent.com/1186748/82619073-c79cfa80-9b89-11ea-9a14-294a8984819d.png)

the .kodiak.toml

```
version = 1
[merge]
automerge_label = "auto-merge" # default: "automerge"
require_automerge_label = true
blacklist_title_regex = "" # default: "^WIP:.*", options: "" (disables regex), a regex string (e.g. ".*DONT\s*MERGE.*")
blacklist_labels = [] # default: [], options: list of label names (e.g. ["wip"])
method = "squash" # default: "merge", options: "merge", "squash", "rebase"
delete_branch_on_merge = true # default: false
block_on_reviews_requested = false # default: false
notify_on_conflict = true # default: true
optimistic_updates = true # default: true
dont_wait_on_status_checks = [] # default: [], options: list of check names (e.g. ["ci/circleci: lint_api"])
update_branch_immediately = false # default: false
prioritize_ready_to_merge = false # default: false
do_not_merge = false # default: false

[merge.message]
title = "github_default" # default: "github_default", options: "github_default", "pull_request_title"
body = "github_default" # default: "github_default", options: "github_default", "pull_request_body", "empty"
include_pr_number = true # default: true
body_type = "markdown" # default: "markdown", options: "plain_text", "markdown", "html"
strip_html_comments = false # default: false
```

I would expect that kodiakhq:status check to show completed. Please let me know what other info I can provide. Thanks for your efforts on this project.

sbdchd:
Just to make sure I understand,  Kodiak is merging PRs but after merge, the status message on github is showing `status ⛴ attempting to merge PR (merging)` instead of something like `merge complete`?

Hmm seems like we don't set a status message on merge success: https://github.com/chdsbd/kodiak/runs/691144867

haghabozorgi:
[...]

PR #469, opened 2020-07-18
add(bot): setting success status on merge complete

sbdchd:
Previously we never update the status after merging a PR which means
the status stays at `status ⛴ attempting to merge PR (merging)`.

Now we show `merge complete 🎉`, but I'm open to other messages / emojis.

example of previous behavior: https://github.com/chdsbd/kodiak/runs/691144867

fixes: https://github.com/chdsbd/kodiak/issues/401

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -928,6 +928,8 @@ branch protection requirements.
                 )
             )
             await api.create_comment(body)
+        else:
+            await set_status("merge complete 🎉")
 
     else:
         position_in_queue = await api.queue_for_merge()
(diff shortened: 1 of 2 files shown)


bot/kodiak/evaluation.py
```diff
@@ -928,6 +928,8 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
                 )
             )
             await api.create_comment(body)
+        else:
+            await set_status("merge complete 🎉")
```
chdsbd, line 932: I'm not sure it matters, but if an existing PR is already merged we will not set this status message. Maybe that's a good thing because that's fewer API calls wasted.

PR #472, opened 2020-07-24
update(bot): evaluation timeout to reduce errors

chdsbd:
I think it was a mistake to not handle these timeout errors. If we hit a timeout Kodiak's basically dead until it receives another message for that pull request.

I think a better solution is to queue the pull request again on timeout. This way we won't loose the pull request and it won't block others because it will go at the end of the queue.

I'm also bumping the timeout because with our api rate limit we've been hitting a lot of timeouts recently.

FYI, we don't have any test coverage on this code.

_This pull request is best viewed with whitespace changes hidden_

### DIFF
--- a/bot/kodiak/pull_request.py
+++ b/bot/kodiak/pull_request.py
@@ -83,7 +83,7 @@ async def evaluate_pr(
                 requeue_callback=requeue_callback,
                 queue_for_merge_callback=queue_for_merge_callback,
             ),
-            timeout=10,
+            timeout=30,
         )
         if pr is None:
             log.info("failed to get_pr")
@@ -113,7 +113,7 @@ async def evaluate_pr(
                     api_call_retry_timeout=api_call_retry_timeout,
                     api_call_retry_method_name=api_call_retry_method_name,
                 ),
-                timeout=10,
+                timeout=30,
             )
             log.info("evaluate_pr successful")
         except RetryForSkippableChecks:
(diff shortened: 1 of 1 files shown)


sbdchd:
How hard would it be to add some smoke tests?

bot/kodiak/pull_request.py
```diff
@@ -73,69 +73,74 @@ async def evaluate_pr(
...
+                # can be resolved by calling GitHub again.
+                if api_call_retry_timeout:
+                    api_call_retry_method_name = e.method
+                    api_call_retry_timeout -= 1
+                    log.info("problem contacting remote api. retrying")
+                    continue
+                log.exception("api_call_retry_timeout")
+            return
+        except asyncio.TimeoutError:
+            # On timeout we add the PR to the back of the queue to try again.
```
sbdchd, line 141: would this be sufficient to handle the timeouts at 10 seconds, aka if we just do this do we need to bump the timeouts on the api requests?

chdsbd: I'm not sure what the timeout should be. Originally I added it to prevent tasks from running indefinitely, which I was suspicious about. https://github.com/chdsbd/kodiak/pull/414

We throttle the api requests we sent to GItHub to be within the 5000/hour rate limit. So some of this wait time could be waiting for that throttle.

chdsbd:
I think we'd need to mock a lot to be able to write a test for this and I'm not enthusiastic about doing that.

chdsbd:
I deployed this change, resolved the Sentry reports, and did a smoke test. I wasn't able to trigger the timeout issue so we'll see how this change fares with more production traffic.

PR #474, opened 2020-07-27
add(web): better subscription exemption support

chdsbd:
This change adds an easier way to mark an account as exempt from subscriptions.

GitHub Sponsors, Enterprise Plan customers, non profits, etc, do not require subscriptions. The current solution is to set a trial that expires in 10 years, but that's a hack.

This change adds a flag we can set in the database to exempt an account and copy that we can set to provide a custom message.

## examples

GitHub user account view (before)
image

GitHub user account view (after)
image

GitHub organization account view (unchanged)
image

GitHub organization account subscribed view (unchanged)
image

Exempted GitHub organization with default message (new)
image

Exempted GitHub organization with custom message (new)
image


web_api/core/test_views.py
```diff
@@ -380,6 +380,31 @@ def test_usage_limit_billing_access_to_owners_member(
     assert res.json()["subscription"]["limitBillingAccessToOwners"] is True
 
 
+@pytest.mark.django_db
+def test_usage_billing_subscription_exemption(
+    authed_client: Client, user: User
+) -> None:
```
sbdchd, line 386: doc comment?

PR #483, opened 2020-08-10
fix(bot): trigger test commit loop

chdsbd:
When a pull request is merged the `mergeable` status becomes `UNKNOWN`. This meant we had an infinite loop where we'd trigger a test commit, queue the PR for another evaluation and continue to trigger test commits.

The fix is to only trigger test commits on open PRs.

Related #482

### DIFF
--- a/bot/kodiak/test_evaluation.py
+++ b/bot/kodiak/test_evaluation.py
@@ -1646,6 +1646,9 @@ async def test_mergeable_pull_request_need_test_commit_need_update_pr_not_open()
     """
     If a pull request mergeable status is UNKNOWN _and_ it is OPEN we should
     trigger a test commit and queue it for evaluation.
+
+    Regression test to prevent infinite loop calling trigger_test_commit on
+    merged/closed pull requests.
     """
     api = create_api()
     mergeable = create_mergeable()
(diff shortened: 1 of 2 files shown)


bot/kodiak/test_evaluation.py
```diff
@@ -1627,6 +1627,7 @@ async def test_mergeable_pull_request_need_test_commit_need_update() -> None:
     pull_request.title = "WIP: add(api): endpoint for checking notifications"
 
     pull_request.mergeable = MergeableState.UNKNOWN
+    pull_request.state = PullRequestState.OPEN
```
chdsbd, line 1630: this is the default value but I thought I'd be explicit.

bot/kodiak/test_evaluation.py
```diff
@@ -1640,6 +1641,46 @@ async def test_mergeable_pull_request_need_test_commit_need_update() -> None:
     assert api.queue_for_merge.called is False
 
 
+@pytest.mark.asyncio
+async def test_mergeable_pull_request_need_test_commit_need_update_pr_not_open() -> None:
+    """
+    If a pull request mergeable status is UNKNOWN _and_ it is OPEN we should
+    trigger a test commit and queue it for evaluation.
```
sbdchd, line 1648: note that this is a regression test meant to catch infinite loops?

Issue #493, opened 2020-08-19
Tag not replicated in Docker hub after release v0.20.0


etiennetremel:
It seems that Git tags aren't correctly pushed to Docker hub. The last available tag is v0.20.0.
Still, I can use the commit sha but having back the tag there would be better for user experience.
Anything I can help with to get the tag available there as well as in git?

chdsbd:
I think we need to update the `.circleci/config.yml` to build tags.

https://circleci.com/docs/2.0/configuration-reference/#tags

PR #495, opened 2020-08-20
tools(ci): trigger docker build on git tag creation

etiennetremel:
The [circleci documentation](https://circleci.com/docs/2.0/configuration-reference/#filters-1) is not super clear, let's try this configuration. If the jobs doesn't get trigger anymore when change are made in a branch, we could add this additional snippet:
```
filters:
  branches:
    only: /.*/
  tags:
    only: /^v\d*\.\d*\.\d*$/
```

fix: https://github.com/chdsbd/kodiak/issues/493

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -335,7 +335,10 @@ workflows:
     jobs:
       - bot_test
       - bot_lint
-      - bot_build_container
+      - bot_build_container:
+          filters:
+            tags:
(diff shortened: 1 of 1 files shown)


etiennetremel:
It seems that those build_container steps should be splitted into to, one for build which run all the time and one for push which only run on master change/git tag creation. Should I make these change part of this PR?

Issue #521, opened 2020-10-09
Feature Request: Can kodiak use multiple automerge_label?


styfle:
### Problem

We are thinking of changing our PRs to require labels to merge.

Rather than adding a new bot, I was wondering if Kodiak could enforce this requirement.

### Possible Solutions

Use an array: `merge.automerge_label = ["semver-major", "semver-minor", "semver-patch"]`

Use a regex pattern: `merge.automerge_label_pattern = "^semver-(.*)$"`

Must match all patterns: `merge.automerge_label_pattern = ["^semver-(.*)$", "^type-(.*)$"]`

styfle:
I started this as a question and then changed to a feature request 😄

chdsbd:
This change hasn't been documented or released yet, but I think it may be what you want: https://github.com/chdsbd/kodiak/pull/516

Although looking at your suggestions it might be less confusing if we have `merge.automerge_label` accept an array as you suggested instead of adding the `automerge_labels` field.

styfle:
> Although looking at your suggestions it might be less confusing if we have merge.automerge_label accept an array as you suggested instead of adding the automerge_labels field.

Either way seems fine to me 👍 

I'm particularly fond of the last solution I suggested because it allows for requiring 2 labels (or more) which opens up really nicely labelled PRs.

```
merge.automerge_label_pattern = ["^semver-(.*)$", "^type-(.*)$"]
```

This example would require both the semver prefixed label and the type prefixed label.

chdsbd:
Do you foresee needing multiple patterns?

Also, do you think you'll need anything to do semver comparisons, like only matching on minor version changes?

styfle:
> Do you foresee needing multiple patterns?

Perhaps in the future yes.

The idea is to avoid PR Templates with checkboxes because that data is not easily searched/filtered compared to labels.

> Also, do you think you'll need anything to do semver comparisons, like only matching on minor version changes?

I didn't think about it at first but now that you bring it up, maybe kodiak could increase the number of reviewers required for semver-major 😮 

That part is not necessary thought. The purpose of the semver label is to better organize release notes and also filter PRs to show all semver-major in the previous release in case there was an unexpected regression.

chdsbd:
Would you want to require every PR to have these labels? If so, I think a GitHub action to enforce the label requirements might be helpful.

styfle:
> Would you want to require every PR to have these labels

Yes that is the goal: labels required to merge. 

[...]

PR #516, opened 2020-10-02
add(bot): merge.automerge_labels config option

chdsbd:
Add new config option (`merge.automerge_labels`) to allow specifying additional automerge_labels.

related: https://github.com/chdsbd/kodiak/issues/476#issuecomment-702077444

PR #522, opened 2020-10-09
ref(bot): make merge.automerge_label accept arrays

chdsbd:
I think it will be less confusing for users if we have a single automerge label field instead of two.

`merge.automerge_labels` hasn't been released yet, so it's safe to remove. After seeing some ideas for using multiple labels with `merge.automerge_label` https://github.com/chdsbd/kodiak/issues/521#issue-718325968, I think it's better if we accept either a string or union of strings.

related #516

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -564,7 +564,7 @@ async def mergeable(
 
     pull_request_labels = set(pull_request.labels)
     config_automerge_labels = (
-        set([config.merge.automerge_label])
+        {config.merge.automerge_label}
         if isinstance(config.merge.automerge_label, str)
         else set(config.merge.automerge_label)
     )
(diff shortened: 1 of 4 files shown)


bot/kodiak/evaluation.py
```diff
@@ -563,8 +563,10 @@ async def set_status(msg: str, markdown_content: Optional[str] = None) -> None:
             return
 
     pull_request_labels = set(pull_request.labels)
-    config_automerge_labels = set(
-        config.merge.automerge_labels + [config.merge.automerge_label]
+    config_automerge_labels = (
+        set([config.merge.automerge_label])
```
sbdchd, line 567:
```suggestion
        {config.merge.automerge_label}
```

sbdchd (approved):

chdsbd:
This change has been released with https://github.com/chdsbd/kodiak/releases/tag/v0.30.0.

PR #553, opened 2020-11-06
ref(bot): log repo info w/ fail in fetch default branch name

sbdchd:
Currently we only log the message and the response which doesn't give us
much to go off.

Instead we can use the logger with the repo name and org already bound.

### DIFF
--- a/bot/kodiak/queries.py
+++ b/bot/kodiak/queries.py
@@ -792,7 +792,7 @@ class Client:
         data = res.get("data")
         errors = res.get("errors")
         if errors is not None or data is None:
-            logger.error("could not fetch default branch name", res=res)
+            self.log.error("could not fetch default branch name", res=res)
             return None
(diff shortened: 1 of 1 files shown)


bot/kodiak/queries.py
```diff
@@ -792,7 +792,7 @@ async def get_default_branch_name(self) -> Optional[str]:
         data = res.get("data")
         errors = res.get("errors")
         if errors is not None or data is None:
-            logger.error("could not fetch default branch name", res=res)
+            self.log.error("could not fetch default branch name", res=res)
```
sbdchd, line 795: we configure this in the init method, should just work:

https://github.com/chdsbd/kodiak/blob/2ecca4ed181432b192292084719e3796d9997812/bot/kodiak/queries.py#L746-L748

chdsbd (approved):

PR #557, opened 2020-11-11
ref(bot,web_api): upgrade deps to get types, remove unnecessary deps

sbdchd:
Upgrade pytest-mock and inflection to get type stubs.

Also removed `ujson` and `arrow` since we can use the stdlib instead.


bot/kodiak/test_queries.py
```diff
@@ -170,32 +170,32 @@ def block_event() -> EventInfoResponse:
         ],
         reviews=[
             PRReview(
-                createdAt=arrow.get("2019-05-22T15:29:34Z").datetime,
+                createdAt=datetime.fromisoformat("2019-05-22T15:29:34+00:00"),
```
sbdchd, line 173: fromisoformat() won't parse `Z` for some reason

chdsbd (approved):

PR #559, opened 2020-11-12
add(mypy): type stubs for some ignored libraries

sbdchd:
Added basic type stubs for some of the smaller libraries the bot and
web_api use.

We put the types in the `typings` folder which we configure in the Mypy, 
but is also the [default path that pyright looks at](https://github.com/microsoft/pyright/blob/b9e61e9f2ae6179284fed8e7b534e5dcb46c8f8b/docs/configuration.md)
so pylance should just work.

Libraries with missing types that remain:
- structlog, waiting on the release of its typed api
- requests_async, need to convert to httpx which has types
- asyncio_redis, maybe write stubs, or use a different client
- django, stubs exist, last time we tried it, there were a number of
  false positives
- stripe, no types currently, could probably write some minimal stubs
  rel: https://github.com/stripe/stripe-python/issues/650

rel: https://github.com/indygreg/python-zstandard/issues/120
rel: https://github.com/davidblewett/rure-python/issues/23
rel: https://github.com/jacobian/dj-database-url/issues/135


bot/typings/rure.pyi
```diff
@@ -0,0 +1,12 @@
+from re import Match, RegexFlag
+from typing import AnyStr, Optional, Pattern, overload
+
+_FlagsType = RegexFlag
```
chdsbd, line 4: why do you assign here instead of using RegexFlag directly?

chdsbd (approved):

bot/typings/rure.pyi
sbdchd: idk, I just styled this after type shed

PR #560, opened 2020-11-14
add(ci): skip build if no changes

chdsbd:
The GitHub bot, docs, web ui, and web api are isolated in their own folders and do not depend on anything outside of those folders.

Assuming master is always green, we only need to run tests in a specific folder when there is a difference against master.

With this change we can reduce our CI time by only running tests for modified folders. Updating the docs does not require testing the GitHub bot.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,9 @@ jobs:
           password: $DOCKER_PASS
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


chdsbd:
This isn't safe because if the docker image changed for our CI, the tests would be skipped when they need to actually run.

PR #564, opened 2020-11-20
add(bot): type stubs for asyncio_redis library

sbdchd:
asyncio_redis does some [dynamic stuff with `__getattr__`][0]
such that the redis commands are defined in protocol, not connection.

For our intents and purposes it's okay to lie and put them on the
Connection class.

[0]: https://github.com/chdsbd/asyncio-redis/blob/319c3aa80f40eccd730ae72e52064874d4428551/asyncio_redis/connection.py#L145-L151


bot/kodiak/queries.py
```diff
@@ -1129,7 +1129,9 @@ async def get_subscription(self) -> Optional[Subscription]:
             # to be backwards compatible we must handle the case of `data` missing.
             try:
                 subscription_blocker = SeatsExceeded.parse_raw(
-                    real_response.get(b"data")
+                    # Pydantic says it doesn't allow Nones, but passing a None
+                    # raises a ValidationError which is fine.
+                    real_response.get(b"data")  # type: ignore
```
sbdchd, line 1134: could also index and catch the `KeyError` down below

chdsbd (approved):

Issue #556, opened 2020-11-10
Queue positions per target branch


staabm:
We see unnecessary queueing of merges, which lead to wait-times.
In our case it’s unnecessary because we see PR B and PR C wait for PR A, even though B and C have a different target branch then A.
Even when all checks on PR B and C are green, they still wait for PR A (which in our case takes 30-45 minutes longer to complete its checks)

Wouldn‘t it make sense to use separate queues per target branch?

Our config: https://raw.githubusercontent.com/redaxo/redaxo/master/.github/.kodiak.toml

chdsbd:
You're right, there's only one merge queue. It definitely makes more sense to have a queue per target branch.

This is where we currently queue pull requests.
https://github.com/chdsbd/kodiak/blob/ec209b151b8a4dc4c9b0c36ae5cb12d9419feb52/bot/kodiak/queue.py#L273-L290

I'm not 100% certain, but I think if we update this function to return a queue name based on the target branch, that should give us per-branch queues. We'd need to lookup the target branch for the pull request to make this work.

https://github.com/chdsbd/kodiak/blob/ec209b151b8a4dc4c9b0c36ae5cb12d9419feb52/bot/kodiak/queue.py#L308-L309

chdsbd:
#572 has been deployed which should resolve this issue. Let me know if you run into any issues.

PR #572, opened 2020-11-23
add(bot): queue per target branch

chdsbd:
We now have a merge queue per repo, per target branch. Previously we only had one merge queue for repo, so PRs would wait for other branches unnecessarily.

On deployment we should delete the `kodiak_merge_queue_names` key to clear out invalid merge queue names.

related #556


bot/kodiak/queue.py
```diff
@@ -315,7 +316,7 @@ async def enqueue_for_repo(
 
 
 def get_merge_queue_name(event: WebhookEvent) -> str:
```
chdsbd, line 318: We start a queue workers per queue name, so by updating this queue name to be per target, we'll get queues per target branch.

bot/kodiak/event_handlers.py
```diff
@@ -101,7 +103,7 @@ async def status_event(status_event: StatusEvent) -> None:
                 api_client.get_open_pull_requests(head=f"{owner}:{ref}") for ref in refs
             ]
             pr_results = cast(
-                List[Optional[List[GetOpenPullRequestsResponse]]],
+                List[Optional[List[GetOpenPullRequestsResponseSchema]]],
```
sbdchd, line 106: I think this cast can be removed since mypy will infer it already

bot/kodiak/queue.py
sbdchd: Could we try and get the old version and the new version at the same time?

chdsbd: I'm not sure what you mean.

bot/kodiak/queue.py
```diff
@@ -18,7 +19,7 @@
 logger = structlog.get_logger()
 
 
-MERGE_QUEUE_NAMES = "kodiak_merge_queue_names"
+MERGE_QUEUE_NAMES = "kodiak_merge_queue_names:v2"
```
chdsbd, line 22: With this change we ignore the existing merge queues on start since they old names are no longer used.

bot/kodiak/queue.py
sbdchd: I was thinking we'd check for existence of the old queues, before creating the new ones

Not sure if that would work though, because when would the repo get switched to the new queue?

sbdchd (approved):

Issue #555, opened 2020-11-10
Feature Request: Multiple configurations for different labels


csprle:
Is it possible to have multiple merge configurations for different labels with Kodiak?
The problem is, for specific pull requests we want to have them merged at prio1 without having to look at other PRs that may have the `automerge` label on them and are waiting in queue.
For all other pull requests, we want the standard queue.

One solution could be adding a second config for another label:

```
[merge]
automerge_label = "automerge"

[merge]
automerge_label = "fast-lane"
prioritize_ready_to_merge = true
```

Another solution that comes to my mind is having the `prioritize_ready_to_merge` to not be a boolean only but also checking for a different label/array of labels.

```
[merge]
automerge_label = "automerge"
prioritize_ready_to_merge_label = "fast-lane"
```

What do you think? Thank you for your help.

chdsbd:
I think your second suggestion is easier to implement. Basically we want to add a label that will allow a PR to skip the merge queue.

We'd need to update this code that checks for `merge.prioritize_ready_to_merge` to also check for the existence of the `merge.prioritize_ready_to_merge_label` label on the pull request.

https://github.com/chdsbd/kodiak/blob/ec209b151b8a4dc4c9b0c36ae5cb12d9419feb52/bot/kodiak/evaluation.py#L995-L1003

chdsbd:
I think the config option should probably be called something like `merge.priority_merge_label`.

I think we have two options for implementing this feature.

We could disable the merge queue to prevent a race between the merge queue updating and merging a pull request while we try to update and merge our priority pull request.

A simpler idea is to place our priority pull request at the front of the merge queue. If there was an existing pull request being merged, we'd have to wait for it.

@csprle Would placing the pull request at the front of the merge queue would meet your need?

csprle:
Yes, placing it at the front of the merge queue is what we need. No need to disable the whole queue.

chdsbd:
#573 has been deployed which should resolve this issue. Let me know if you run into any issues.

PR #573, opened 2020-11-23
add(bot): merge.priority_merge_label

chdsbd:
Add new config option to place PR at front of merge queue. When the `merge.priority_merge_label` label is applied to a pull request, the pull request will be placed at the front of the merge queue.

If two PRs have the `merge.priority_merge_label` label, either one may be first in queue.

related #555


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- add new config option (`merge.priority_merge_label`) to place PR at front of merge queue. (#573, #555)
```
chdsbd, line 11: I think we should get in the habit of writing to the CHANGELOG when we add code instead of when we deploy.

bot/kodiak/queue.py
```diff
@@ -143,7 +143,7 @@ async def requeue() -> None:
             only_if_not_exists=True,
         )
 
-    async def queue_for_merge() -> Optional[int]:
+    async def queue_for_merge(*, first: bool) -> Optional[int]:
```
sbdchd, line 146: why does this function exist?

chdsbd: Because `evaluate_pr` requires it, but when we're merging a PR we will never call it.

sbdchd (approved):

PR #574, opened 2020-11-26
add(web_api): django type stubs

sbdchd:
Add some basic django type stubs. There are some ignores around
`Request.user` since we don't inherit from `AbstractBaseUser`.

We're on an older version of Django that doesn't support indexing
into `QuerySet` or `BaseManager`, so we monkey patch those.

### DIFF
--- a/web_api/web_api/patches.py
+++ b/web_api/web_api/patches.py
@@ -1,0 +1,9 @@
+from __future__ import annotations
+
+from django.db.models.manager import BaseManager
+from django.db.models.query import QuerySet
+
+
+def patch_django() -> None:
+    for cls in (QuerySet, BaseManager):
+        cls.__class_getitem__ = classmethod(lambda cls, *args, **kwargs: cls)  # type: ignore [attr-defined]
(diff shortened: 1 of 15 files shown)


web_api/web_api/patches.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+from django.db.models.manager import BaseManager
+from django.db.models.query import QuerySet
+
+
+def patch_django() -> None:
```
chdsbd, line 7: maybe add a doc comment explaining why we're doing this?

web_api/web_api/views.py
```diff
@@ -681,8 +686,8 @@ def process_login_request(request: HttpRequest) -> Union[Success, Error]:
     # handle errors
     if request.POST.get("error"):
         return Error(
-            error=request.POST.get("error"),
-            error_description=request.POST.get("error_description"),
+            error=request.POST["error"],
+            error_description=request.POST["error_description"],
```
chdsbd, line 690: Just to ensure this never breaks
```suggestion
            error=request.POST["error"],
            error_description=request.POST.get("error_description") or "",
```

chdsbd (approved):

PR #575, opened 2020-11-26
add(bot,web_api): require generic arguments

sbdchd:
Add most (all?) the mypy options to the bot and web_api config and more
importantly enable `disallow_any_generics` so that we can't accidentally
forget a generic argument.


bot/kodiak/app_config.py
```diff
@@ -12,7 +12,7 @@
 
 
 class TypedConfig(Config):
-    @overload  # type: ignore [override]
+    @overload
```
sbdchd, line 15: not really sure what caused this change

chdsbd (approved):

Issue #577, opened 2020-11-28
Too many blank lines before co-authors if message body is empty


gharlan:
If the message body is empty (because pr description is empty or `body` is configured as `empty`) there are too many blank lines between title and co-authors.

Example commit message:

```
PR title (#issue)



Co-authored-by: ...
```

chdsbd:
To match [the GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) we add two new lines before the coauthor trailers.

<img width="692" alt="Screen Shot 2020-11-28 at 1 01 07 PM" src="https://user-images.githubusercontent.com/1929960/100522826-ca505c00-3179-11eb-8c69-ca0e14d57fd3.png">

It seems that GitHub adds an extra space between the commit title and body. I think we can fix the issue you've encountered by trimming any extra whitespace on merge.

chdsbd:
#578 has been deployed which should resolve this issue. Let me know if you run into any issues.

PR #578, opened 2020-11-28
add(bot): limit coauthor trailer whitespace

chdsbd:
If the commit message body is empty, don't add the two line spacing before coauthor trailers. GitHub adds it's own line between the title and body, so we'll have a three line space between the title and trailers if we don't remove our added spacing.

related #577

### DIFF
--- a/bot/kodiak/test_evaluation.py
+++ b/bot/kodiak/test_evaluation.py
@@ -3294,7 +3294,16 @@ async def test_mergeable_include_coauthors() -> None:
     config = create_config()
     config.merge.message.include_coauthors = True
 
-    for body_style in (MergeBodyStyle.pull_request_body, MergeBodyStyle.empty):
+    for body_style, commit_message in (
+        (
+            MergeBodyStyle.pull_request_body,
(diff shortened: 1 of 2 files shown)


bot/kodiak/test_evaluation.py
```diff
@@ -3294,7 +3294,16 @@ async def test_mergeable_include_coauthors() -> None:
     config = create_config()
     config.merge.message.include_coauthors = True
 
-    for body_style in (MergeBodyStyle.pull_request_body, MergeBodyStyle.empty):
+    for body_style, commit_message in (
+        (
+            MergeBodyStyle.pull_request_body,
+            "# some description\n\nCo-authored-by: Barry Block <73213123+b-block@users.noreply.github.com>",
```
sbdchd, line 3300: I think we should have more tests with some tabs and other whitespace in the commit message

chdsbd: how about now that we're not trimming anything.

sbdchd (approved):

PR #579, opened 2020-11-28
ref(bot): move get_commit_authors into separate file

chdsbd:
I think long term we should break up queries.py into multiple fields. I find it confusing having all the pydantic schemas in the same file.

### DIFF
--- a/bot/kodiak/test_queries.py
+++ b/bot/kodiak/test_queries.py
@@ -674,8 +674,8 @@ def test_get_commit_authors_error_handling() -> None:
     pull_request_data = {
         "commitHistory": {
             "nodes": [
-                {"commit": {"author": {"user": {}}}},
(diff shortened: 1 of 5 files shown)


sbdchd (approved):

bot/kodiak/test_queries.py
```diff
@@ -668,8 +674,8 @@ def test_get_commit_authors_error_handling() -> None:
     pull_request_data = {
         "commitHistory": {
             "nodes": [
-                {"commit": {"author": {"user": {}}}},
-                {"commit": {"author": {}}},
+                {"commit": {"author": {"user": None}}},
+                {"commit": {"author": None}},
```
chdsbd, line 678: This didn't parse and I don't think the GitHub API will return responses like this.

PR #580, opened 2020-11-29
ref(bot): move commit author logic into evaluation.py

chdsbd:
This change supports a future change to ignore merge commits. de-duping commit authors is now in evaluation.py where most of our app logic lives.


bot/kodiak/queries/commits.py
```diff
@@ -51,8 +50,9 @@ def get_commit_authors(*, pr: Dict[str, Any]) -> List[User]:
...
     if not nodes:
         return []
+    commit_authors = []
     for node in nodes:
         if node.commit.author is None or node.commit.author.user is None:
             continue
-        commit_authors[node.commit.author.user] = True
-    return list(commit_authors.keys())
+        commit_authors.append(node.commit.author.user)
+    return commit_authors
```
sbdchd, line 58: seems like you're changing this from what is essentially a set to a list

chdsbd: yeah, I realized it's better to do this in `evaluation.py` because we also need to handle the pull request author case where we want to add the pull request author, but not add a duplicate trailer.

sbdchd (approved):

PR #592, opened 2020-12-10
dep(bot): update markdown-html-finder to version with Python 3.8 support

chdsbd:


chdsbd:
fixed by uploading more Wheels for the library

Issue #606, opened 2021-01-22
auto merge dependency updates based on upgrade type


chdsbd:
As a user I want Kodiak to auto merge "patch" and "minor" version bumps, but I want to review "major" version upgrades myself.

chdsbd:
This was resolved in #607. The Kodiak docs describe the feature in detail: https://kodiakhq.com/docs/recipes#configuring-automerge-by-upgrade-type

PR #607, opened 2021-01-28
add(bot): auto merge by dependabot version

chdsbd:
Add configuration option to auto merge PRs opened by Dependabot (and others) based on the upgrade type (major, minor, patch).

Snyk creates batch PRs, but Dependabot doesn't. I think for a first pass we can just support single dependency pull requests.

```toml
[merge]
automerge_label = "ship it!"

[merge.automerge_dependencies]
versions = ["minor","patch"]
usernames = ["dependabot"]

[update]
ignored_usernames = ["dependabot"]
```

related #606

### DIFF
--- a/bot/kodiak/config.py
+++ b/bot/kodiak/config.py
@@ -55,7 +55,7 @@ DEFAULT_TITLE_REGEX = "^WIP:.*"
 
 class AutomergeDependencies(BaseModel):
     versions: List[Literal["major", "minor", "patch"]] = []
-    usernames: List[str] = ["snyk-bot", "dependabot"]
+    usernames: List[str] = []
 
 
 class Merge(BaseModel):
--- a/bot/kodiak/config.py
+++ b/bot/kodiak/config.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Optional, Union, cast
 
 import toml
 from pydantic import BaseModel, ValidationError, validator
+from typing_extensions import Literal
 
 
 class MergeMethod(str, Enum):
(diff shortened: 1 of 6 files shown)


bot/kodiak/config.py
```diff
@@ -52,9 +53,15 @@ class MergeMessage(BaseModel):
 DEFAULT_TITLE_REGEX = "^WIP:.*"
 
 
+class AutomergeDependencies(BaseModel):
+    versions: List[Literal["major", "minor", "patch"]] = []
+    usernames: List[str] = ["snyk-bot", "dependabot"]
```
sbdchd, line 58: shouldn't this be disabled by default aka empty list?

bot/kodiak/dependencies.py
```diff
@@ -0,0 +1,83 @@
...
+    old_version_parts = _parse_version_simple(old_version)
+    new_version_parts = _parse_version_simple(new_version)
+
+    for part in new_version_parts[:3] + old_version_parts[:3]:
+        try:
+            int(part)
+        except ValueError:
+            return None
+
+    if _get_or_none(new_version_parts, 0) != _get_or_none(old_version_parts, 0):
```
sbdchd, line 66: if these are both None it will return "major"

bot/kodiak/dependencies.py
```diff
@@ -0,0 +1,83 @@
...
+) -> Optional[Literal["major", "minor", "patch"]]:
+    """
+    Determine patch, like Dependabot.
+
+    https://github.com/dependabot/dependabot-core/blob/998de3be7811956354aea077ecb180831e24012c/common/lib/dependabot/pull_request_creator/labeler.rb#L92-L114
+    """
+    old_version_parts = _parse_version_simple(old_version)
+    new_version_parts = _parse_version_simple(new_version)
+
+    for part in new_version_parts[:3] + old_version_parts[:3]:
```
sbdchd, line 60: why `:3`?

bot/kodiak/dependencies.py
chdsbd: Wouldn't it not match, so it would return patch?

```python
assert (None != None) is False
```

bot/kodiak/dependencies.py
chdsbd: Because that's what the dependabot code does and it returns the "major", "minor", "patch"-like versions if available.

```
4.1.2 => ["4","1","2"]

2018.1.2.523.b1 => ["2018","1","2"]
```

bot/kodiak/dependencies.py
sbdchd: maybe it should be, if either of them are `None` make it a `patch`

Thoughts?

sbdchd (approved):

Issue #624, opened 2021-03-15
Potential Bug Risks and Anti-Patterns


HarshCasper:
## Description 

Hi 👋 

I ran DeepSource Static Code Analysis upon the Project, the results for which are available [here](https://deepsource.io/gh/HarshCasper/kodiak/). 

The Static Code Analysis Tool found potential bugs and anti-patterns in the Code, that can be detrimental at a later point in time with respect to the Project. DeepSource helps you to automatically find and fix issues in your code during code reviews. This tool looks for anti-patterns, bug risks, performance problems, and raises issues. 

Some of the notable issues are: 

- Parameters mismatch in overridden method [here](https://deepsource.io/gh/HarshCasper/kodiak/issue/PYL-W0221/occurrences)
- Missing props validation in a React component definition [here](https://deepsource.io/gh/HarshCasper/kodiak/issue/JS-0463/occurrences)
- Unused arguments in the functions [here](https://deepsource.io/gh/HarshCasper/kodiak/issue/PYL-W0613/occurrences)
- Unnecessary fragments [here](https://deepsource.io/gh/HarshCasper/kodiak/issue/JS-0424/occurrences)

There are plenty of other issues in relation to Bug Discovery and Anti-Patterns which you would be interested to take a look at.

If you would like to integrate DeepSource to autofix some of the commonly occurring issues, I can help you set that up :)

sbdchd:
Thanks for linting the repo!

Your service has a slick UI but I think I prefer sticking to ESLint, Flake8, etc. rather than using a SaaS.

Also the number of false positives is too many, which makes it hard to use. Pylint has a similar issue where it will warn about all sorts of stuff by default and requires a lengthy config.

- Parameters mismatch in overridden method -- ignored via `type: ignore [override]`
- Missing props validation in a React component definition -- since we're using typescript no need for the runtime type assertions
- Unused arguments in the functions -- a few valid unused arguments, with a number of false positives
- Unnecessary fragments -- looks neat, we should setup https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md

PR #626, opened 2021-03-15
add(web_ui): eslint for useless jsx fragments

sbdchd:
surfaced in https://github.com/chdsbd/kodiak/issues/624

see: https://github.com/yannickcr/eslint-plugin-react/blob/eeb92731bd922af4493cf938f0547a021e5cdbcf/docs/rules/jsx-no-useless-fragment.md

### DIFF
--- a/web_ui/src/components/SideBarNav.tsx
+++ b/web_ui/src/components/SideBarNav.tsx
@@ -148,7 +148,7 @@ function Loading() {
   return (
     <SideBarNavContainer
       orgContent={<SkeletonProfileImage />}
-      switchAccountContent={<></>}
+      switchAccountContent={null}
     />
   )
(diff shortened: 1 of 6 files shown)


chdsbd (dismissed):

web_ui/src/components/SideBarNav.tsx
```diff
@@ -148,7 +148,7 @@ function Loading() {
   return (
     <SideBarNavContainer
       orgContent={<SkeletonProfileImage />}
-      switchAccountContent={<></>}
```
sbdchd, line 151: arguably this is a false positive, but `null` is more obvious than `<></>`

chdsbd (approved):

PR #628, opened 2021-03-18
fix(bot): support graceful container shutdown

chdsbd:


bot/Dockerfile
```diff
@@ -21,4 +21,4 @@ RUN rm -rf /var/app/pip-wheel-metadata/
 # install cli
 RUN poetry install
 
-CMD /var/app/.venv/bin/uvicorn kodiak.main:app --host 0.0.0.0 --port $PORT
+CMD /var/app/.venv/bin/python -m kodiak.main
```
sbdchd, line 24: doesn't this need to use the array syntax like `CMD ["binary", "arg1", "arg2"]` otherwise it will spawn a shell?

sbdchd (approved):

PR #649, opened 2021-04-12
add paging support to queries.get_open_pull_requests

wuub:
as it currently stands this method returns only a default page size
of 30 pull requests.  for many repositories a number of open PRs can
easily reach hundreds, all often targeting the same branch

this change bumps the default page_size to 100 and traverses pages
until either an empty one is reached OR a limit of ~200~ 20 pages happens
first (as a failsafe)

I briefly considered using `link` header with rel=prev, rel=last as
a loop condition, but it looked fragile


wuub:
sorry for the ammends 

I have not noticed many of the circleci checks locally until I issued a PR
I had to go trough a few iterations to make the mypy and black gods happy :)

chdsbd:
@wuub Thanks for the PR.

Increasing the page size to 100 sounds good to me.

One caveat is that by increasing the page size, Kodiak will be slightly slower for installations with more than thirty pull requests. GitHub Apps are limited to ~1.5 API requests per second. So more API requests make Kodiak slower for a given installation.

Out of curiosity, have you run into the 30 PR limit with Kodiak? Did you encounter this with `push` events or `status` events?

bot/kodiak/queries/__init__.py
```diff
@@ -993,18 +993,39 @@ async def get_open_pull_requests(
             params["base"] = base
         if head is not None:
             params["head"] = head
+
+        params["per_page"] = "100"
+        open_prs = []
+
         async with self.throttler:
```
chdsbd, line 1000: We need to wrap each API call (`self.session.get`) with this context manager.

So it would be:

```python
while page:
    with self.throttler:
        res = await self.session.get(...)
```

wuub: on it.

wuub:
> Out of curiosity, have you run into the 30 PR limit with Kodiak? Did you encounter this with push events or status events?

I noticed it on a repo with high traffic and around ~100 PRs open at any given time. It's using a non-standard, non-merging kodiak config. With the primary intention of using `autoupdate` label to keep the PR up do date w/ `main` branch (even at the cost of increased CI jobs)

```
version = 1

[merge]
automerge_label = "automerge"
do_not_merge = true

[update]
autoupdate_label = "autoupdate"
require_automerge_label = false 
```

The issue is, that w/o paging it misses those few PRs w/ an `autoupdate` label regularly. They only get updated when I touch them in some other way (additional coment, or a unrelated label)

wuub:
I addressed the comment about throttler usage and also added a test which checks the PR limit path (so it's indeed 2000 and not 20000)

chdsbd:
This PR looks good. I made some minor changes to your branch. I'll merge this PR after I can verify the change on a staging environment.

I realized that the new tests are super slow (>40s) because we run the real throttler in tests. I'll make a different PR to patch out the throttler in tests.

Unrelated, is the repository with Kodiak updating a bunch of branches public? If so, do you have a link?

Also, do you use the Kodiak GitHub Marketplace App or do you self host?

wuub:
> I made some minor changes to your branch.

sure thing :) i'm new to the code base (and to modern'ish python as well TBH) it's expected I'll make mistakes

> I realized that the new tests are super slow (>40s) because we run the real throttler in tests. I'll make a different PR to patch out the throttler in tests.

🙇 

> Unrelated, is the repository with Kodiak updating a bunch of branches public? If so, do you have a link?

Unfortunately no, the repo in question is super-duper-private. FWIW the `get_open_pull_requests` enabled image has been running flawlessly on it for the last ~24hrs. I woke up to almost 50 mails about kodiak-bot merging `main` branch into my test PRs

> Also, do you use the Kodiak GitHub Marketplace App or do you self host?

For now I self host. 
The repositories I'm testing kodiak bot on are not "normal" (think hundreds of stale PRs ). I'm expecting to unearth more things like `get_open_pull_requests` and I need a way to test any fixes in the field before I send them your way. 
 
Eventually I would love to run upstream unchanged or even better, move to the SasS model, but I'm not there yet.

chdsbd (approved):
Thanks for the contribution!

PR #651, opened 2021-04-14
fix(bot): patch throttler in tests to make queries faster

chdsbd:
Throttling our API calls in tests makes them run slowly. This change patches the throttler so we don't wait.

I created a ThrottlerProtocol to verify that our fake throttler matches the real throttle api we use.

related: #649


bot/kodiak/tests/fixtures.py
```diff
@@ -19,3 +19,11 @@ def create_commit(
             )
         ),
     )
+
+
+class FakeThottler:
```
sbdchd, line 24: I think you can inherit from the protocol and the type checker will enforce that the shape matches

chdsbd: does work :(

sbdchd (approved):

PR #656, opened 2021-05-13
remove redis dependency in test_queries.test_get_event_info_blocked

wuub:
when testing locally test_queries.test_get_event_info_blocked blocks
indefinitely w/o a local redis instance running

while I don't fully understand if the real redis instance was needed, the fake_redis approach passes the tests, so 🤷‍♂️

### DIFF
--- a/bot/kodiak/test_queries.py
+++ b/bot/kodiak/test_queries.py
@@ -3,12 +3,10 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, Iterable, cast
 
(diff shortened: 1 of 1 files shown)


chdsbd:
@wuub Thanks for the PR. Hanging tests are annoying.

My personal preference is to not mock out Redis, so I've made a PR (#657) that should fix this issue by skipping the test if Redis isn't running

wuub:
reasonable :) 
thanks @chdsbd

PR #657, opened 2021-05-13
fix(bot): skip redis-dependent test if Redis isn't running

chdsbd:
fixes #656


sbdchd (approved):
surprised only one test requires redis

PR #659, opened 2021-06-04
ref(bot): incoming webhook processing

sbdchd:
Right now the incoming webhook HTTP server is kind of mixed in with the
event processing code.

This is an issue when it comes to deploy code & with the daily restarts
because we can't stop only the event processing code, instead we stop 
the HTTP as well which causes us to drop events -- not good.

New plan, we'll have a couple entrypoints:

- HTTP server handling incoming webhooks which does as little processing
  as possible before dumping the events onto a queue
- Queue consumers, the existing logic would work off the new incoming
  event queue instead of in the HTTP server's endpoint handler.

We'll probably want a way to run these two entry points together so
people self hosting don't have issues.


bot/kodiak/queue.py
```diff
@@ -334,4 +334,10 @@ def get_webhook_queue_name(event: WebhookEvent) -> str:
...
 
 
-redis_webhook_queue = RedisWebhookQueue()
+async def enqueue_raw_webhook(
+    *, redis: asyncio_redis.Connection, event_name: str, event: dict[str, object]
+) -> None:
+    logger.info("enqueuing github webhook event")
+
+    # TODO(sbdchd): how do we want to format the queue messages?
+    await redis.rpush(event)
```
sbdchd, line 343: @chdsbd how do we want to format the incoming raw events from GitHub?

chdsbd: Maybe something like this. I think we need to record the event name, which is passed in via an HTTP header. We can probably dump the payload body as a string from JSON
```py
class RawEvent:
    name: str
    payload: str
```

bot/kodiak/redis.py
```diff
@@ -0,0 +1,33 @@
...
+        try:
+            redis_db = int(conf.REDIS_URL.database)
+        except ValueError:
+            redis_db = 0
+        conn = await asyncio_redis.Pool.create(
+            host=conf.REDIS_URL.hostname or "localhost",
+            port=conf.REDIS_URL.port or 6379,
+            password=conf.REDIS_URL.password or None,
+            db=redis_db,
+            poolsize=conf.REDIS_POOL_SIZE,
```
sbdchd, line 30: I think since this is only going to be used for putting incoming messages on the queue, we may need a different redis pool size env var

bot/kodiak/main.py
```diff
@@ -103,5 +106,16 @@ async def webhook_event(
     await enqueue_incoming_webhook(redis=redis, event_name=github_event, event=event)
 
 
+def main() -> None:
+    # TODO(sbdchd): we should make this an argument and do something proper like arg parse
```
chdsbd, line 110: we have click installed

bot/kodiak/cli.py
```diff
@@ -93,3 +94,42 @@ def refresh_pull_requests() -> None:
...
+    from kodiak.queue import main as queue_consumers_main
+
+    http_process = Process(target=http_main)
+    http_process.start()
+
+    queue_consumers_process = Process(target=queue_consumers_main)
+    queue_consumers_process.start()
+
+    http_process.join()
+    queue_consumers_process.join()
```
chdsbd, line 135: If queue_consumers_process dies before http_process, what happens? Don't we want to exit immediately in that case?

bot/kodiak/redis.py
chdsbd: we definitely do

docs/docs/self-hosting.md
```diff
@@ -101,3 +101,9 @@ These instructions describe setting up Kodiak on Heroku using a Docker container
     # release app
     heroku container:release web -a $APP_NAME
     ```
+
+    > **NOTE:** While the default CMD for the Docker container will run Kodiak,
+    > for a more robust setup you can run the webhook http server and the queue
+    > consumers seperately. Entrypoints should be `/var/app/.venv/bin/kodiak http-server` and `/var/app/.venv/bin/kodiak queue-consumers` using the same
+    > Docker container image. This allows for updating & restarting Kodiak with
+    > less/zero downtime.
```
chdsbd, line 109:
```suggestion
    > **NOTE:** By default, Docker will run Kodiak's http server and queue consumers together.
    > For a more robust deployment, you can run the webhook http server and the queue
    > consumers seperately. Entrypoints should be `/var/app/.venv/bin/kodiak http-server` and `/var/app/.venv/bin/kodiak queue-consumers` using the
    > Kodiak Docker image. This allows for updating & restarting the Kodiak queue workers with
    > less downtime.
```

bot/kodiak/redis.py
sbdchd: hmm, we use this `get_conn` function in both the http server and the queue consumers, thoughts?

bot/kodiak/cli.py
sbdchd: sounds good

bot/kodiak/redis.py
sbdchd: I've add a param so they can specify

chdsbd (approved):

PR #664, opened 2021-06-05
cut(bot): unused param valid_signature

sbdchd:
This param is passed down into the evaluation logic, but isn't used.

We validate webhook signatures before putting them on the queue.

`git log -S` points to this param never being used.


chdsbd (approved):
I think the purpose of this check was to handle the `requiresCommitSignatures` branch protection rule.

This code is obviously broken. I just tested and we hit the generic "🛑 cannot merge (Merging blocked by GitHub requirements)" message if a commit isn't signed.

We can always add a more detailed error message in the future if someone asks

PR #669, opened 2021-06-07
ref(queue-consumers): combine event_handler and queue modules

sbdchd:
In refactoring the incoming consumer to be separate from the HTTP server
there ends up being a circular dependency between these two modules.

For now we can combine them and figure out a better structure later.


bot/kodiak/events/__init__.py
```diff
@@ -1,16 +1,5 @@
-from typing import Mapping, Type
-
-from kodiak.events.base import GithubEvent
-from kodiak.events.check_run import CheckRunEvent
-from kodiak.events.pull_request import PullRequestEvent
-from kodiak.events.pull_request_review import PullRequestReviewEvent
-from kodiak.events.push import PushEvent
-from kodiak.events.status import StatusEvent
-
-event_schema_mapping: Mapping[str, Type[GithubEvent]] = {
```
sbdchd, line 10: this mapping thing is unused

chdsbd (approved):

PR #683, opened 2021-07-01
add(docs): logos to homepage

chdsbd:
It seems like other websites to not make their customer logos links, which I think makes sense since it reduces the chance of someone clicking away. I've added title attributes to all the logos with the website name.

All the logos are the same height. I tried to use SVGs when available. Surprisingly, Docusaurus _does not support importing react components_. We need v2 for that, so all the logos are in the index.js file.

This is the first batch of logos. I'll add more in different PRs.


sbdchd (approved):
Might try one of those SVG compression tools to reduce the size a bit

PR #709, opened 2021-08-13
fix(web_api): billing payment method updates

chdsbd:
Previously if a user updated their payment method, their old payment method would still be attached to the subscription. This basically caused renewals to error until there was manual intervention.

Now, when a new payment method is attached, we remove the old ones so there is only ever one payment method attached to a customer.

We also use the first payment method available on an account for sending invoices when a user adds seats. Previously we would use the payment method attached to a subscription, but when we delete a user's initial payment method, that subscription.default_payment_method becomes null.


sbdchd (approved):

sbdchd:
Got some lint & test errors fyi

PR #715, opened 2021-08-24
feat(doc): mention renovate support

chdsbd:
I didn't want to duplicate every reference for Dependabot, so I sprinkled around a few mentions of Renovate.


sbdchd (approved):
LGTM

PR #723, opened 2021-09-13
deps: update uvicorn to 0.15.0

jbergstroem:
..alongside its dependencies.

This change was introduced by invoking `poetry update uvicorn` after updating `pyproject.toml`.

Hoping that CI will cover more test setups than mine.

### DIFF
--- a/bot/poetry.lock
+++ b/bot/poetry.lock
@@ -1,10 +1,42 @@
 [[package]]
+name = "anyio"
+version = "3.3.1"
+description = "High level compatibility layer for multiple asynchronous event loop implementations"
+category = "main"
+optional = false
+python-versions = ">=3.6.2"
+
+[package.dependencies]
+idna = ">=2.8"
+sniffio = ">=1.1"
+typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
+
+[package.extras]
+doc = ["sphinx-rtd-theme", "sphinx-autodoc-typehints (>=1.2.0)"]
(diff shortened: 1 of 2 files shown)


jbergstroem:
So, pip fails with:
```
Creating virtualenv kodiak-py3.7 in /root/project/bot/.venv
Installing dependencies from lock file
                                  
[NonExistentKey]   
'Key "hashes" does not exist.'  
                                  
install [--no-dev] [--dry-run] [-E|--extras EXTRAS] [--develop DEVELOP]
```

According to google, this is likely due to using an old version of poetry. Should I update that as well in this PR?

chdsbd:
@jbergstroem Yeah, we're on an old version of poetry ([0.12.12](https://github.com/chdsbd/kodiak/blob/4c4ff6c2665796f640c5d32b013e042a6d0112b2/.circleci/config.yml#L31)). I'm not opposed to upgrading our poetry version if you're up for it. We need to update `bot/` and `web_api/` as well as the CircleCI config.

jbergstroem:
> @jbergstroem Yeah, we're on an old version of poetry (`0.12.12`). I'm not opposed to upgrading our poetry version if you're up for it. We need to update bot/ and web_api/ as well as the CircleCI config.

Will do a new PR later today to keep things separate.

jbergstroem:
(will rebase once poetry lands)

jbergstroem:
Rebased but fails fast in CI; looks like old poetry is still being pulled. The circleci cache documentation is easy to read; so is the config. Even the cache step says "no cache". What am I missing? Is it related to force pushing to the branch?

chdsbd:
@jbergstroem The PR looks good to me. It seems like it's passing CI at the moment

chdsbd (approved):

PR #729, opened 2021-09-17
support org level config in $owner/.github repo

### LINKED ISSUE #403: Support organisation-wide configuration via <org>/.github repositories
I'd love to use Kodiak across all my repos with the same configuration, but it can be very time consuming to port this file across all repos and make sure they don't drift apart.

GitHub already uses `<org>/.github` repos for configuration across all repos, including issue templates, so I think that could be a good place to keep that global config file. [See this example](https://github.com/relaycorp/.github/).

### LINKED ISSUE #703: Can Kodiak work without config file in the repo?
We'd like to use the tool internally for ~50 different repositories and we'd prefer to configure Kodiak once on the server side and do not modify repositories. Would it be possible?

LeoQuote:
fix #403
fix #703 

could need additional permission for contents


chdsbd:
This looks great @LeoQuote! I've been a little busy, but I will merge this PR when I get a chance to test it

chdsbd:
I test this and made some changes so I think it should work now

LeoQuote:
good work!

chdsbd (approved):

Issue #671, opened 2021-06-10
merge.message.body could be a template string


jbergstroem:
It would be interesting to allow further customization of the message body by providing a simple template (python template string? mustache? plain variables?) that allows for minor customization. I would for one like to add metadata to the body such as committer, reviewer or author and a link to the PR URL (See the [node.js project for an example](https://github.com/nodejs/node/commit/ff09d6de166bf5ea85e3decb723ab3915662e719)):

Example:
```toml

[merge.message]

body = """
$pr_body

PR-URL: $pr_url"""

chdsbd:
Thanks for the suggestion.

Kodiak has partial support for some of the metadata you'd like to record.

You can enable [`merge.message.include_pull_request_url`](https://kodiakhq.com/docs/config-reference#mergemessageinclude_pull_request_url) to add the pull request URL to the commit body, but the URL wouldn't have the `PR-URL:` prefix. Is that something you could live without? Maybe we could update Kodiak to include that prefix to better match the `Co-authored-by` trailers.

You can use [`merge.message.include_coauthors`](https://kodiakhq.com/docs/config-reference#mergemessageinclude_coauthors) to add `Co-authored-by` trailers to the PR.

Kodiak doesn't have anything to help with `Reviewed-By` trailers, but it probably wouldn't be hard to support.

If possible, my preference would be to avoid adding template support, but I'm open to the idea if we need to do it

jbergstroem:
> Kodiak has partial support for some of the metadata you'd like to record.

Excellent!

> Kodiak doesn't have anything to help with Reviewed-By trailers, but it probably wouldn't be hard to support.

Right. Consider me a user if supported (as well as the "missing `PR-URL`").

> If possible, my preference would be to avoid adding template support, but I'm open to the idea if we need to do it

It would be useful for me to understand your position here; why you think templates is not the path forward. Nitpick, but lets say someone dislike your order? Wants to add emojis (I hear thats a thing now); just overall more flow control if you get me.

chdsbd:
@jbergstroem I'm okay with adding templating, but if we can add more boolean options instead I'd prefer that. My thought is that boolean config options are more foolproof, although less flexible, than templating.


This is where Kodiak sets the `Co-authored-by` trailers and the PR url: https://github.com/chdsbd/kodiak/blob/c786e9f1e95aee0a67468b86815c023086c1608f/bot/kodiak/evaluation.py#L201-L242

[...]

PR #730, opened 2021-09-22
feat: prefix URLs with `PR-URL`

jbergstroem:
When enabling `include_pull_request_url` (defaults to false), prefix the URL added to the commit body with `PR-URL`. This follows other trailer syntax such as `Co-Authored-By:` as well as the general recommendations by [conventional commits][1] footer best practices.

Not sure how to handle with the fact that this could be seen as SEMVER breaking; changing expectations of commit body and all..

References: https://github.com/chdsbd/kodiak/issues/671#issuecomment-858240918

[1]: https://www.conventionalcommits.org/en/v1.0.0/

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -201,11 +201,6 @@ def get_merge_body(
         MergeBodyStyle.pull_request_body,
         MergeBodyStyle.empty,
     ):
-        if config.merge.message.include_pull_request_url:
-            if merge_body.commit_message is None:
-                merge_body.commit_message = pull_request.url
-            else:
-                merge_body.commit_message += "\n\n" + pull_request.url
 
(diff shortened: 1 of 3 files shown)


jbergstroem:
(the `typing-extensions` extension seems to have issues getting installed)

chdsbd:
@jbergstroem Poetry is being a pain like normal. I'm going to try to resolve that and then I'll merge this in.

I'm fine with this change even if it is technically a breaking change

jbergstroem:
> @jbergstroem Poetry is being a pain like normal. I'm going to try to resolve that and then I'll merge this in.

I noticed. Happy to do some lifting here too; getting acquainted with it.

> I'm fine with this change even if it is technically a breaking change

Ok! I will go after `Reviewed-By` next.

sbdchd (approved):
LGTM

PR #731, opened 2021-09-22
fix: bot and web_api build

chdsbd:
The Poetry install for purest was failing because it's dependency, pluggy was failing to install because it's dependency, [zipp was failing to install](https://app.circleci.com/pipelines/github/chdsbd/kodiak/2434/workflows/08bc1ae3-77d4-4057-a961-badbc4f5b4e5/jobs/21448).

Bumping the pytest version fixes the install, but I also needed to upgrade pytest-asyncio to fix a deprecation error.


sbdchd (approved):
I think the API pyproject file also needs to be updated

sbdchd:
Looks like we might need to upgrade our poetry version to avoid that runtime bug: https://github.com/python-poetry/poetry/issues/4409#issuecomment-903173358

chdsbd:
#732 should resolve this

PR #737, opened 2021-09-29
fix(bot): httpx instability by downgrading httpcore

### LINKED ISSUE #736: latest httpcore upgrade breaks merging
After our recent dependency upgrades, we now see occasional, [`NewConnectionRequired`](https://github.com/encode/httpcore/blob/c8d93d07369dab11d5c03f11de25607be1d21faf/httpcore/_async/base.py#L8-L9) exceptions being raised.

If this exception is raised during the merge flow, Kodiak will break out of merging the current PR and move on to the next. 

https://github.com/chdsbd/kodiak/blob/7b8243b385e9c0bd1462f111b8b88af59e91d28a/bot/kodiak/queue.py#L409

Until https://github.com/encode/httpcore/issues/335 is fixed, we should revert to our old dependency version.

chdsbd:
Our recent dependency upgrades moved us to a new version of httpcore that has been raising a NewConnectionRequired exception, breaking merging for some users.

I think this commit is the problem: https://github.com/encode/httpcore/commit/98fb2a0814d1fd076a354804f071cdcef9812615#diff-4f89d2e45c858632692840675cf189f73a499f51d39fc25c8f5bac828bd735e3R138

fixes #736

### DIFF
--- a/bot/poetry.lock
+++ b/bot/poetry.lock
@@ -1,22 +1,4 @@
 [[package]]
-name = "anyio"
-version = "3.3.1"
-description = "High level compatibility layer for multiple asynchronous event loop implementations"
-category = "main"
-optional = false
-python-versions = ">=3.6.2"
-
-[package.dependencies]
-idna = ">=2.8"
-sniffio = ">=1.1"
-typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
-
-[package.extras]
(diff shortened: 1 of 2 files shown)


chdsbd:
I've already deployed this change to production btw

sbdchd (approved):
Yikes

PR #738, opened 2021-09-30
fix(bot): handle rure errors

chdsbd:
It's possible to provide invalid regex patterns to rure, which will trigger an exception. This change handles that exception and displays a user-facing error.


sbdchd (approved):

sbdchd (commented):
I think the type stubs may need to be updated: 

```
kodiak/evaluation.py:774:12: error: Module has no attribute "exceptions"  [attr-defined]
```

PR #740, opened 2021-10-04
fix(bot): GraphQL API call failures breaking merge loop

### LINKED ISSUE #739: better handle 502 errors when merging PRs
When Kodiak starts merging a PR, it enters a loop, polling until the PR is ready to merge. If Kodiak makes a GitHub API request and encounters an error, it will eject from this loop, ignoring the PR and moving on to the next PR in queue.

This is problematic behavior as it can trigger extra updates to PRs. We should gracefully retry on recoverable errors.

Here's where we check for an error:
https://github.com/chdsbd/kodiak/blob/afd6e59833f95ec4a28487f95fb09b3a515fd1c6/bot/kodiak/queries/__init__.py#L876-L880

We should audit all of our API requests to ensure we gracefully handle errors.

chdsbd:
fixes #739

### DIFF
--- a/bot/kodiak/pull_request.py
+++ b/bot/kodiak/pull_request.py
@@ -147,6 +147,7 @@ async def evaluate_pr(
             return
         except asyncio.TimeoutError:
             # On timeout we add the PR to the back of the queue to try again.
+            # We're in a while loop, so we won't loose our spot if we're merging.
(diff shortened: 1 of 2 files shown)


chdsbd:
I've deployed this change to production

PR #749, opened 2021-11-10
feat(bot): use single ingest queue per installation

chdsbd:
Previously we had a shared pool of ingest queue workers. This is bad because a lot of workers can get tied up waiting on the API rate limit throttle.

Now we start an ingest worker per installation.


bot/kodiak/entrypoints/worker.py
```diff
@@ -75,6 +123,14 @@ async def main() -> NoReturn:
                 queue.start_webhook_worker(queue_name=task_meta.queue_name)
             else:
                 assert_never(task_meta.kind)
+        if ingest_queue_watcher.done():
+            logger.info("worker task failed", kind="ingest_queue_watcher")
+            exception = ingest_queue_watcher.exception()
```
sbdchd, line 128: shouldn't we check if this exception is non `None` before logging / sending off to sentry?

chdsbd: it shouldn't ever finish

Issue #753, opened 2021-11-17
Redis password doesn't seem to work anymore after upgrade 0.45 > 0.47 upgrade 


etiennetremel:
I'm using a self-hosted Kodiak and after upgrading from 0.45 to 0.47 Kodiak cannot connect to Redis using password authentication.
I'm using a Redis URL with the following format: `REDIS_URL=redis://:secret123@localhost:6379`.

Looking at the diff between both versions, I'm seeing [the following line](https://github.com/chdsbd/kodiak/blob/master/bot/kodiak/entrypoints/ingest.py#L44-L46) which append `.encode()` to the password variable.
But if I use the same code locally and remove the `.encode()` it works fine. 

Was there a specific reason to use `.encode()`, is it necessary or am I missing something?

chdsbd:
Oh, I think that's a mistake.

I think I copied the code from the queue.py, which uses the BytesEncoder and requires converting the string to bytes.

https://github.com/chdsbd/kodiak/blob/9dce1bccf95b48afa552ec179477a47eaae7dc83/bot/kodiak/entrypoints/ingest.py#L44-L46

I think we can remove the `.encode()` from the following three places

https://github.com/chdsbd/kodiak/blob/9a8ca6e68e6493455b3fe4977270fdee6b4a14dc/bot/kodiak/entrypoints/ingest.py#L45
https://github.com/chdsbd/kodiak/blob/9a8ca6e68e6493455b3fe4977270fdee6b4a14dc/bot/kodiak/redis_client.py#L17
https://github.com/chdsbd/kodiak/blob/38db2d2097001dfbec44f6f31722c37c84801d26/bot/kodiak/test_queries.py#L270

etiennetremel:
Thanks for the quick reply and review @chdsbd !

PR #754, opened 2021-11-17
fix: remove .encode() from redis password

etiennetremel:
close https://github.com/chdsbd/kodiak/issues/753


chdsbd:
Thanks for the PR!

chdsbd (approved):

Issue #770, opened 2021-12-20
feature: option to remove separator as part of cutting body


jbergstroem:
When using [cut body before](https://kodiakhq.com/docs/config-reference#mergemessagecut_body_before) or after, there should be an option to include removing the separator from said result.

At my company we take parts of the PR body, validate it with `commitlint` and if ok, tell kodiak to use it for body. We use the markdown hr separator to split what we want in the commit body and what we can keep in the PR body (`---`). Right now, if I use `cut_body_after`, it keeps the separator:

### Original body
```
Here's a body

---

And here's the footer
```

### Expected output
```
Here's a body

```

### Actual output
```
Here's a body

---
```

I would like to suggest an option that toggles the removal of the separator as part of cutting the body.

Also, perhaps something that removes extra newlines (1 empty newline between title, body and footer seems to be the norm)?

jbergstroem:
> I'm not sure what the new configuration option should be called. Maybe merge.message.cut_body_and_text

I have a PR almost done, chose the name `strip_separator: (True|False)`

jbergstroem:
FYI, ran into issues updating the config schema:
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/jbergstroem/wrk/oss/kodiak/bot/kodiak/cli.py", line 9, in <module>
    from kodiak import app_config as conf
  File "/Users/jbergstroem/wrk/oss/kodiak/bot/kodiak/app_config.py", line 39, in <module>
    "REDISCLOUD_URL", cast=databases.DatabaseURL
  File "/Users/jbergstroem/wrk/oss/kodiak/bot/kodiak/app_config.py", line 32, in __call__
    return super().get(key, cast=cast, default=default)  # type: ignore [arg-type]
[...]

PR #771, opened 2021-12-20
feat(bot): optionally strip the cut separator

jbergstroem:
If `cut_body_after` or `cut_body_before` is in use, setting this option to `True` also strips the separator itself.

~Note, this is a draft; awaiting the following input:~
1. [x] Is there a better name? (renamed to `cut_body_and_text`)
2. [x] Should we improve the way the body string is managed?
3. [x] json schema generation is [having issues](https://github.com/chdsbd/kodiak/issues/770#issuecomment-998293493) with `REDISCLOUD_URL` and `SECRET_KEY`

Fixes: https://github.com/chdsbd/kodiak/issues/770

### DIFF
--- a/bot/kodiak/evaluation.py
+++ b/bot/kodiak/evaluation.py
@@ -74,6 +74,7 @@ def get_body_content(
     strip_html_comments: bool,
     cut_body_before: str,
     cut_body_after: str,
+    cut_body_and_text: bool,
     pull_request: PullRequest,
 ) -> str:
     if body_type is BodyText.markdown:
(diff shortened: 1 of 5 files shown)


bot/kodiak/evaluation.py
```diff
@@ -74,18 +74,24 @@ def get_body_content(
     strip_html_comments: bool,
     cut_body_before: str,
     cut_body_after: str,
+    cut_body_and_text: bool,
     pull_request: PullRequest,
 ) -> str:
     if body_type is BodyText.markdown:
         body = pull_request.body
+        # TODO(jbergstroem): This could probably be made faster by just cutting the string differently
```
chdsbd, line 82: I don't think the performance here will be too much of an issue

chdsbd (approved):
I think this looks great!

bot/kodiak/evaluation.py
jbergstroem: Me neither, just wanted to leave it open for you to decide.

jbergstroem:
@chdsbd could you enable the checks so we can see if it passes?

PR #775, opened 2022-01-10
Update Mergify

jd:
It looks like this comparison is quite old, updating some of the fields.


chdsbd:
Hey @jd, thanks for the PR! I made a couple updates to your change. The date in the comment is the creation date for the service, which we use for sorting the table.

In my opinion, the Mergify config is not simple. Also, I don't really consider Mergify open source because only a piece of the code is available, not everything you need to run Mergify yourself.

chdsbd (approved):

PR #794, opened 2022-03-22
update(web_api): index more created_at based queries

sbdchd:
technically we only need to index:
  - github_event.created_at
  - user_pull_request_activity_progress.created_at
  - pul_request_activity_progress.created_at

but it was easier to update the parent class instead of
the specific ones.


web_api/web_api/migrations/0026_auto_20220322_0036.py
```diff
@@ -0,0 +1,53 @@
...
+
+    dependencies = [
+        ('web_api', '0025_auto_20200902_0052'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='account',
+            name='created_at',
+            field=models.DateTimeField(auto_now_add=True, db_index=True),
```
sbdchd, line 16: do these need to use concurrent stuff or is it nbd?

chdsbd: It's probably okay if we don't

chdsbd (approved):

PR #796, opened 2022-03-28
chore(bot): change Docker base image to python:3.7-slim

etiennetremel:
While scanning the Kodiak Docker image with [Grype](https://github.com/anchore/grype), we noticed a significant amount of vulnerabilities.

This PR upgrade the base image to the latest 3.7 image and use the slim version which come with a smaller size and reduce the attack surface. I also took the opportunity to use a non root user.

- previously: image size 1.31GB, 3694 vulnerabilities
- now: image size 607Mb, 87 vulnerabilities

### DIFF
--- a/bot/Dockerfile
+++ b/bot/Dockerfile
@@ -1,32 +1,32 @@
-FROM python:3.7@sha256:6eaf19442c358afc24834a6b17a3728a45c129de7703d8583392a138ecbdb092
+FROM python:3.7-slim
 
-RUN set -ex && mkdir -p /var/app
-
-RUN apt-get update && apt-get install -y supervisor
-
-RUN mkdir -p /var/log/supervisor
-
-# use cryptography version for poetry that doesn't require Rust
-RUN python3 -m pip install cryptography===37.0.4
-RUN python3 -m pip install poetry===1.1.13
-
-RUN poetry config virtualenvs.in-project true
-
-COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+RUN apt update && \
+  apt-get install --no-install-recommends -y \
+    supervisor \
+    git && \
+  python -m pip install --upgrade pip && \
+  pip install \
+    --no-cache-dir \
+    --root-user-action=ignore \
+    cryptography===37.0.4 \
+    poetry===1.1.15 && \
+  poetry config virtualenvs.in-project true && \
+  groupadd kodiak && \
+  useradd --uid 1000 --gid kodiak kodiak && \
+  mkdir -p /var/app && \
+  chown -R kodiak:kodiak /var/app
 
 WORKDIR /var/app
 
-COPY pyproject.toml poetry.lock /var/app/
+COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+
+COPY --chown=kodiak pyproject.toml poetry.lock /var/app
 
 # install deps
 RUN poetry install
 
-COPY . /var/app
+COPY --chown=kodiak . /var/app
 
-# workaround for: https://github.com/sdispater/poetry/issues/1123
-RUN rm -rf /var/app/pip-wheel-metadata/
-
-# install cli
-RUN poetry install
+USER kodiak
 
 CMD ["/usr/bin/supervisord"]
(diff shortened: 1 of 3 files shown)


sbdchd:
Nice, I love that smaller image size!

etiennetremel:
Anything else I can help with?

chdsbd:
@etiennetremel It looks the CI job for building the bot container failed. Should be good to merge once that's fixed

etiennetremel:
ah, that's interesting.. it worked on my machine 😛
I made a few changes, let's see how that goes

etiennetremel:
@chdsbd anychance you could run the docker build job with credentials? 
image

bot/Dockerfile
```diff
@@ -1,30 +1,27 @@
...
-RUN set -ex && mkdir -p /var/app
-
-RUN apt-get update && apt-get install -y supervisor
-
-RUN mkdir -p /var/log/supervisor
-
-RUN python3 -m pip install poetry===1.1.13
-
-RUN poetry config virtualenvs.in-project true
-
```
sbdchd, line 12: Don’t we still need the supervisor conf?

etiennetremel: Good catch, I just added that missing line and also updated supervisord.conf to redirect stderr logs to stdout. It makes it easier to debug.

bot/Dockerfile
```diff
@@ -1,32 +1,30 @@
-FROM python:3.7@sha256:6eaf19442c358afc24834a6b17a3728a45c129de7703d8583392a138ecbdb092
+FROM python:3.7-slim@sha256:e720c669cd7188d6a7e36e567be6bde9feb931d66219f1e4ba2caeb30e35d2cd
```
sbdchd, line 1: looks good, one minor thing is we should probably have the tests run in a python:3.7-slim container instead of the current python:3.7 to head off any differences before release 

Also @chdsbd probably wants to run these containers on staging before merge although I don't think we use any OS specific stuff

etiennetremel: good point, I just updated the container to use slim version for the bot test step in the circleci config

chdsbd:
Not sure why CI is failing

etiennetremel:
> Not sure why CI is failing

Error was being triggered during the pip install of poetry, for this use case I reckon we can safely use the flag `--root-ignore-action=ignore` as we then use kodiak as user to run the app from supervisord.

sbdchd:
Oh I think we need to update CI to install `git` for the script to pass:

https://app.circleci.com/pipelines/github/chdsbd/kodiak/2726/workflows/6fe5f10b-3fc1-4ecd-a1ea-faf4347e28c1/jobs/25425

novascreen:
@etiennetremel @sbdchd any change you could give this another look? it would be really nice to get vulnerabilities resolved

etiennetremel:
@novascreen it's pretty old but I just rebased the branch. The Docker credentials are missing in CircleCI, I reckon only @sbdchd or @chdsbd would be able to help with it.

bot/Dockerfile
```diff
@@ -1,32 +1,32 @@
...
-
-RUN poetry config virtualenvs.in-project true
+WORKDIR /var/app
 
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 
-WORKDIR /var/app
-
-COPY pyproject.toml poetry.lock /var/app/
+COPY --chown=kodiak pyproject.toml poetry.lock /var/app
```
chdsbd, line 23: This appears to error in CI:

```
Step 5/9 : COPY --chown=kodiak pyproject.toml poetry.lock /var/app
When using COPY with more than one source file, the destination must be a directory and end with a /

Exited with code exit status 1
```

https://app.circleci.com/pipelines/github/chdsbd/kodiak/2818/workflows/f4a433c9-18a4-4a29-9873-7e601d2e5a83/jobs/26706

etiennetremel: ahh, good point, I just updated it

chdsbd (approved):
Thank you!

novascreen:
Awesome, thank you!

Issue #800, opened 2022-04-14
Kodiak not stripping HTML comments from PR messages


jpalmieri:
We're seeing html comments stay in the merge messages when merging PRs, despite our configuration matching the recipe for [Better Merge Messages](https://kodiakhq.com/docs/recipes#better-merge-messages) in the docs.

Let me know if there's other information that I could provide which might be relevant to this.

chdsbd:
Can you provide an example of the Kodiak configuration and the pull request body?

jpalmieri:
I'd rather not provide the entire config, but here's the section that seems relevant. If there are other parts that might be relevant, let me know and I can post those.

```
[merge.message]
# by default, github uses the first commit title for the PR of a merge.
# "pull_request_title" uses the PR title.
title = "pull_request_title" # default: "github_default", options: "github_default", "pull_request_title"

# by default, GithHub combines the titles a PR's commits to create the body
# text of a merge. "pull_request_body" uses the content of the PR to generate
# the body content while "empty" simple gives an empty string.
body = "pull_request_body" # default: "github_default", options: "github_default", "pull_request_body", "empty"

# GitHub adds the PR number to the title of merges created through the UI.
# This setting replicates that feature.
include_pr_number = true # default: true

# markdown is the normal format for GitHub merges
body_type = "markdown" # default: "markdown", options: "plain_text", "markdown", "html"

# useful for stripping HTML comments created by PR templates when the `markdown` `body_type` is used.
strip_html_comments = true # default: false
```

I'd rather not send over an actual commit message, but here's an example of what we're seeing 
```
Add cool new feature (#123)

### :label: Jira ticket

<!-- Add the Jira ticket corresponding to this Pull request -->

https://company.atlassian.net/browse/COOL-123

### :loudspeaker: Type of change

<!--- Remove what's irrelevant -->

- New feature (non-breaking change which adds functionality)
### :scroll: Description

<!--
   Describe your changes in detail.
   Mainly answer the "What" and "How". What is the PR changing? What is it adding/modifying/removing? What will the new behavior be?
   How is the change introduced?
-->

Add feature which is really cool and useful.

...
```

chdsbd:
@jpalmieri Thanks! That's enough to replicate the issue on my side. I'll look into this.

Oddly Kodiak did remove the HTML comment in this example PR I used earlier:

```
Bumps [serialize-error](https://github.com/sindresorhus/serialize-error) from 8.1.0 to 9.1.1.
[...]

PR #801, opened 2022-04-14
fix strip html comments

chdsbd:
related #800


chdsbd:
https://github.com/chdsbd/kodiak/issues/800#issuecomment-1099484118

PR #826, opened 2022-07-03
remove FastAPI and just use Starlette

chdsbd:
We don't need FastAPI for our single API endpoint

### DIFF
--- a/bot/kodiak/entrypoints/ingest.py
+++ b/bot/kodiak/entrypoints/ingest.py
@@ -5,15 +5,17 @@ from __future__ import annotations
 
 import hashlib
 import hmac
-from typing import Any, Dict, Optional, cast
+from typing import Any
 
 import asyncio_redis
 import structlog
 import uvicorn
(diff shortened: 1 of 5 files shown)


bot/kodiak/entrypoints/ingest.py
```diff
@@ -52,42 +48,42 @@ async def get_redis() -> asyncio_redis.Pool:
...
+@app.route("/", methods=["GET"])
+async def root(_: Request) -> Response:
+    return PlainTextResponse("OK")
+
+
+@app.route("/api/github/hook", methods=["POST"])
+async def github_webhook_event(request: Request) -> Response:
+    body = await request.body()
     expected_sha = hmac.new(
         key=conf.SECRET_KEY.encode(), msg=(await request.body()), digestmod=hashlib.sha1
```
sbdchd, line 60: Why do we not use the body variable here?

sbdchd (changes requested):
Minor comment

sbdchd (approved):

PR #834, opened 2022-07-09
Run shellcheck in a one-liner

szepeviktor:
Docs: https://www.shellcheck.net/wiki/GitLab-CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -407,7 +407,7 @@ jobs:
       - run:
           name: run shellcheck
           command: |
-            ./s/shellcheck
+            git ls-files --exclude='*.sh' --ignored --cached -z | xargs -0 -r shellcheck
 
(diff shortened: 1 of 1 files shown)


szepeviktor:
If it works the local shell script could be deleted.

.circleci/config.yml
```diff
@@ -407,7 +407,7 @@ jobs:
       - run:
           name: run shellcheck
           command: |
-            ./s/shellcheck
+            git ls-files --exclude='*.sh' --ignored --cached -z | xargs -0 -r shellcheck
```
chdsbd, line 410: We need the s/shellcheck script so we can run this check locally

szepeviktor: Now I have to face that I'm not a developer, and unable to sympathize 😿

szepeviktor: How about moving this one-liner to `s/shellcheck`?

PR #838, opened 2022-07-28
delete subscriptions from db on cancellation

chdsbd:
This helps prevent a Kodiak UI issue where a user still sees a subscription when it's actually cancelled.


sbdchd (approved):
Should we add a log statement or something?

netnimit1998 (commented):
@netnimit

PR #846, opened 2022-10-10
retry when we encounter merge being blocked

chdsbd:
Sometimes we're blocked from merging for an unknown reason. We should retry when merging to prevent queue issues.

I think this problem is a result of some internal GitHub defect.


chdsbd:
I've deployed this change in production

sbdchd (approved):

PR #849, opened 2022-10-18
better support codeowner reviews

chdsbd:
I'll make another PR to block merge for this case once we log this for a bit.


bot/kodiak/test_evaluation.py
```diff
@@ -1538,6 +1538,49 @@ async def test_mergeable_unknown_merge_blockage() -> None:
     assert api.queue_for_merge.called is False
 
 
+@pytest.mark.xfail
+@pytest.mark.asyncio
+async def test_mergeable_unknown_merge_blockage_code_owner() -> None:
```
chdsbd, line 1543: Once we've verified this is correct, we'll fix this test

sbdchd (approved):

PR #852, opened 2022-10-25
fix(bot): http client perf issue

sbdchd:
Work around for a perf issue with how httpx handles SSL.

Instead of using `httpx.AsyncClient` directly, we subclass and reuse the ssl context.

rel: https://github.com/encode/httpx/issues/838

### DIFF
--- a/bot/kodiak/http.py
+++ b/bot/kodiak/http.py
@@ -1,11 +1,10 @@
 from __future__ import annotations
+
 import ssl
 
 from httpx import AsyncClient
-from httpx._types import (
-    TimeoutTypes,
-)
 from httpx._config import DEFAULT_TIMEOUT_CONFIG
+from httpx._types import TimeoutTypes
 
 # NOTE: this has a cost to create so we may want to set this lazily on the first HttpClient creation
 context = ssl.create_default_context()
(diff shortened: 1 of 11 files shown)


bot/kodiak/http.py
```diff
@@ -0,0 +1,29 @@
...
+class HttpClient(AsyncClient):
+    """
+    HTTP Client with the SSL config cached at the module level to avoid perf issues.
+    see: https://github.com/encode/httpx/issues/838
+    """
+
+    def __init__(
+        self,
+        *,
+        timeout: TimeoutTypes = DEFAULT_TIMEOUT_CONFIG,
```
sbdchd, line 23: was thinking we may want to change this since in one place we set it to nothing and the default (5 seconds) might be low

wdyt @chdsbd

bot/kodiak/http.py
```diff
@@ -0,0 +1,28 @@
...
+
+from httpx import AsyncClient
+from httpx._config import DEFAULT_TIMEOUT_CONFIG
+from httpx._types import TimeoutTypes
+
+# NOTE: this has a cost to create so we may want to set this lazily on the first HttpClient creation
+context = ssl.create_default_context()
+
+
+class HttpClient(AsyncClient):
```
sbdchd, line 13: also maybe there's a python lint we can use to ban usage of `AsyncClient`?

sbdchd: done!

chdsbd (approved):

PR #856, opened 2022-11-09
wip(bot): add heartbeat for webhook repo worker

sbdchd:
Basically we want to know if the worker gets stuck and to kill it if it does

We handle it failing due to an exception, but we don't handle cases where it hangs.

### DIFF
--- a/bot/kodiak/queue.py
+++ b/bot/kodiak/queue.py
@@ -481,7 +481,7 @@ class RedisWebhookQueue:
 
     def __init__(self) -> None:
         self.worker_tasks: MutableMapping[
-            str, tuple[Task[NoReturn], Literal["repo", "webhook"], str]
+            str, tuple[Task[NoReturn], Literal["repo", "webhook"]]
         ] = {}  # type: ignore [assignment]
         self.worker_task_heatbeats: MutableMapping[str, int] = {}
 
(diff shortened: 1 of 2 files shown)


bot/kodiak/queue.py
```diff
@@ -546,7 +555,8 @@ def _start_worker(
             sentry_sdk.capture_exception(exception)
         log.info("creating task for queue")
         # create new task for queue
-        self.worker_tasks[key] = (asyncio.create_task(fut), kind)
+        # TODO: before merge fix this
+        self.worker_tasks[key] = (asyncio.create_task(fut), kind, "TODO-id")
```
sbdchd, line 559: we need the task to have access to the worker heartbeat data structure which it needs to update periodically with its status

sbdchd: not sure how to do that yet, may need to restructure some things

bot/kodiak/queue.py
```diff
@@ -481,6 +483,7 @@ def __init__(self) -> None:
         self.worker_tasks: MutableMapping[
             str, tuple[Task[NoReturn], Literal["repo", "webhook"]]
         ] = {}  # type: ignore [assignment]
+        self.worker_task_heatbeats: MutableMapping[str, int] = {}
```
sbdchd, line 486: need to figure out a way for the task to have access to this

sbdchd: I wonder if there is an asyncio way to have the task thrown an exception unless it checks in every so often

PR #867, opened 2023-02-04
replace asyncio-redis with redis-py

chdsbd:
Asyncio-redis is pretty terrible in adverse network conditions. Digital Ocean network connectivity can flake out and the lack of timeouts within asyncio-redis make it easy for the library to cause hangs, which will break Kodiak for users. redis-py's async version (which originally was aioredis) has excellent timeouts for every operation, making it a safer choice. It also has a reasonable connection pool setup where we don't need to have a fixed pool size.

The default timeouts for redis-py async are infinite, which isn't great, but we explicitly set short timeouts.

Hopefully, now we shouldn't see any more hangs, which has been a problem for us for many months.


bot/kodiak/app_config.py
```diff
@@ -55,11 +54,17 @@ def __call__(
         default=["pull_request", "pull_request_review", "pull_request_comment"],
     )
 )
-USAGE_REPORTING_POOL_SIZE = config("USAGE_REPORTING_POOL_SIZE", cast=int, default=50)
 USAGE_REPORTING_QUEUE_LENGTH = config(
     "USAGE_REPORTING_QUEUE_LENGTH", cast=int, default=10_000
 )
 INGEST_QUEUE_LENGTH = config("INGEST_QUEUE_LENGTH", cast=int, default=1_000)
+BLOCKING_POP_TIMEOUT_SEC = config("BLOCKING_POP_TIMEOUT_SEC", cast=int, default=10)
```
sbdchd, line 61: thoughts on calling this `REDIS_BLOCKING_POP_TIMEOUT_SEC`?
```suggestion
REDIS_BLOCKING_POP_TIMEOUT_SEC = config("REDIS_BLOCKING_POP_TIMEOUT_SEC", cast=int, default=10)
```

bot/pyproject.toml
```diff
@@ -9,12 +9,10 @@ python = "^3.7"
 uvicorn = "^0.15.0"
 toml = "^0.10.0"
 PyJWT = "^1.7"
-cryptography = "^3.4.6"
 structlog = "^19.1"
 colorama = "^0.4.1"
 databases = "0.3.2"
 # bzpopmin and bzpopmax support. connection pool is built concurrently. TLS support.
```
sbdchd, line 15:
```suggestion
```

bot/kodiak/refresh_pull_requests.py
```diff
@@ -180,21 +175,19 @@ class RefreshPullRequestsMessage(BaseModel):
...
-        try:
-            res: BlockingPopReply = await redis.blpop(
-                ["kodiak:refresh_pull_requests_for_installation"], timeout=5
-            )
-        except asyncio_redis.exceptions.TimeoutError:
-            logger.info("pull_request_refresh", timeout_reached=True)
+        res = await redis_bot.blpop(
+            ["kodiak:refresh_pull_requests_for_installation"], timeout=5
+        )
+        logger.info("pull_request_refresh", timeout_reached=True)
```
sbdchd, line 182: shouldn't this log be in the line below?

bot/kodiak/redis_client.py
```diff
@@ -1,19 +1,27 @@
-import asyncio_redis
+import redis.asyncio as redis
 
 import kodiak.app_config as conf
 
 
-async def create_connection() -> asyncio_redis.Connection:
+def create_connection() -> 'redis.Redis["bytes"]':
```
sbdchd, line 6:
```suggestion
def create_connection() -> 'redis.Redis[bytes]':
```

bot/kodiak/app_config.py
```diff
@@ -55,11 +54,19 @@ def __call__(
...
 )
-USAGE_REPORTING_POOL_SIZE = config("USAGE_REPORTING_POOL_SIZE", cast=int, default=50)
 USAGE_REPORTING_QUEUE_LENGTH = config(
     "USAGE_REPORTING_QUEUE_LENGTH", cast=int, default=10_000
 )
 INGEST_QUEUE_LENGTH = config("INGEST_QUEUE_LENGTH", cast=int, default=1_000)
+REDIS_BLOCKING_POP_TIMEOUT_SEC = config(
+    "REDIS_BLOCKING_POP_TIMEOUT_SEC", cast=int, default=10
+)
+# if we don't get a reply from Redis within a short period, we have an error because we always expect short response times from redis. We specify a timeout for blocking oeprations
```
sbdchd, line 64:
```suggestion
# if we don't get a reply from Redis within a short period, we have an error because we always expect short response times from redis. We specify a timeout for blocking operations
```

sbdchd (approved):

netnimit1998:
@netnimit1998

PR #881, opened 2023-12-19
Add support for specifing Redis username

### LINKED ISSUE #879: Enable setting username for Redis
I'm trying to set up self-hosted Kodiak, but I'm not able to make it connect to the Redis instance.

My Redis instance is dependent on both username and password, but it seems like this snippet of code is overriding the username and password I set in the URL.

https://github.com/chdsbd/kodiak/blob/729e64dbc1211117a1d83aaccc0bcbd0967e550a/bot/kodiak/redis_client.py#L13-L18

Would be possible to expand this to include the possiblity of setting username as well?

androa:
Fixes #879


chdsbd (approved):
Thanks!

PR #911, opened 2026-03-06
Reduce graphql query cost

chdsbd:
This query change increased the graphql query cost from 1 to 2, reducing Kodiak's throughput


jadouvibeoff-lab:
> This query change increased the graphql query cost from 1 to 2, reducing Kodiak's throughput

